PR #2, opened 2018-10-12
[WIP] Fix missing context on Android

fathyb:
# Wait for segmentio/analytics-android#595 to be released first

Ref: LIB-647


f2prateek:
We could merge this with the RC release tag as the dependency.

pvinis:
@fathyb @f2prateek why was this closed? Can we have it merged? android is still missing custom context from `middleware`.

PR #7, opened 2018-10-22
feat(client): use object-based configuration

fathyb:


README.md
```diff
@@ -94,11 +98,9 @@ In your code :
 import analytics from '@segment/analytics-react-native'
 import GoogleAnalytics from '@segment/analytics-react-native-google-analytics'
 
-await analytics
-  .configure()
-  .using(GoogleAnalytics)
-  // ...
-  .setup('writeKey')
+await analytics.setup('writeKey', {
```
f2prateek: Is this async? Why is the await required?

fathyb: It's required because all calls to the React-Native bridge are async. If the bridge fails for any reason it can be handled here so that we don't have uncaught promise rejections. This is the same behaviour as before, so we need to keep the queuing code.

f2prateek: Got it - why is it async again? Is this a React limitation or something else?

fathyb: Yup, in a production app the thread where the native code is executed is different from the thread where JavaScript is running, so React-Native can only communicate between the two using messages (in iOS they are not even in the same process).

React-Native kinda documents it a bit [here](https://facebook.github.io/react-native/docs/native-modules-ios):
> React Native bridge is asynchronous, so the only way to pass a result to JavaScript is by using callbacks or emitting events (see below).

Also when debugging React-Native, it runs your app in the browser and proxify the bridge over WebSocket (the lets you use the devtools debugger for example). So it should always be assumed that the JS and native part can be run on totally different machines.

f2prateek: Got it - ok!

PR #8, opened 2018-10-22
fix(integrations): fix TypeScript typings

### LINKED ISSUE #6: Branch package not correct typing
Hi, Thanks for this great lib!

I noticed the Branch package isn't the right typing, I get the following error with my code:

```typescript
analytics
  .configure()
    .using(Branch)
    .recordScreenViews()
    .trackAppLifecycleEvents()
    .trackAttributionData()
    .ios()
      .trackAdvertising()
      .trackDeepLinks()
    .android()
      .disableDeviceId()
  .setup('...')
  .then(() => {})
  .catch(() => {});
```

```
Argument of type '{ disabled: true; } | (() => void)' is not assignable to
parameter of type 'Integration'. Type '() => void' is not assignable to
type '    Integration'. Type '() => void' is not assignable
to type '{ disabled: true; }'. Property 'disabled' is missing in type '() => void'.
```

I noticed in the `@segment/analytics-react-native-branch/index.d.ts` file that the branch integration is declared like this:
```typescript
declare const BranchIntegration:
    | {disabled: true}
    | (() => void)
```

Maybe this should be of the `Integration` type to make the typings add up?

A workaround for now is to cast Branch to any (`.using(Branch as any)`).

fathyb:
Fixes #6

### DIFF
--- a/packages/integrations/template/index.d.ts.tpl
+++ b/packages/integrations/template/index.d.ts.tpl
@@ -1,5 +1,5 @@
 declare const {{{slug}}}Integration:
     | {disabled: true}
-    | (() => void)
+    | (() => Promise<void>)
 
 export = {{{slug}}}Integration


fathyb:
Waiting for CircleCI outage to end before merging. (https://status.circleci.com/incidents/02bfh9zqfhp6?u=f0p22lyllfr0)

PR #12, opened 2018-10-24
Pull android tool versions from root project

ajmath:
If the host react native application is using a different version of the android sdk build tools, build warnings will be displayed about incompatibilities.  This addresses that by using the values defined in the root project.

This is the approach used by the current [react-native-create-library template](https://github.com/frostney/react-native-create-library/blob/master/templates/android.js#L28).

### DIFF
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -12,13 +12,17 @@ buildscript {
     }
 }
 
+def safeExtGet(prop, fallback) {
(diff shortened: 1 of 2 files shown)


fathyb:
Thanks! Could you also add this to the integration template?
https://github.com/segmentio/analytics-react-native/blob/7f1985dfa0d7376aba3609cedac71965aa99c74e/packages/integrations/template/android/build.gradle#L15-L25

PR #26, opened 2018-12-04
Swap config comments to match their respective config option.

caroaguilar:
The comments for the `recordScreenViews` and `trackAppLifecycleEvents` were swapped. I just re-ordered the comments so they can match the correct property.


f2prateek:
cc @fathyb

PR #50, opened 2019-03-28
fix(core): React-Native 0.59 Support

### LINKED ISSUE #46: Android Builds failing in RN 0.59
Since upgrading to the recent React Native 0.59, my builds are failing because of this package:

```
FAILURE: Build failed with an exception.

* What went wrong:
The Android Gradle plugin supports only Kotlin Gradle plugin version 1.3.0 and higher.
The following dependencies do not satisfy the required version:
project ':@segment_analytics-react-native' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.2.60```

### PR DESCRIPTION
**Pull request by fathyb**
- React-Native 0.59 support
- Upgrade Detox 9 -> 12
- Upgrade test-app to React-Native 0.59.2
- Commit test-app to git

Fixes #46
Closes #44 

---

Ref: LIB-909

### CODE REVIEW

File: packages/core/android/build.gradle [hunk 1]
```diff
@@ -41,5 +41,5 @@ dependencies {
     api 'com.segment.analytics.android:analytics:4.+'
 
     api 'com.facebook.react:react-native:+'
-    api 'org.jetbrains.kotlin:kotlin-stdlib:1.2.+'
+    api 'org.jetbrains.kotlin:kotlin-stdlib:1.3.+'
```
f2prateek: should we use a specific version rather than latest for these?

fathyb: I think it originally was set by IDEA, do you think it'd be better to not specify any?

f2prateek: I think we should be more specific (e.g. 1.3.2), rather than loose. Otherwise, it might work now but fail later.

packages/core/android/build.gradle
```diff
@@ -41,5 +41,5 @@ dependencies {
     api 'com.segment.analytics.android:analytics:4.+'
 
     api 'com.facebook.react:react-native:+'
```
f2prateek: we should fix these versions too when we get the chance.

PR #54, opened 2019-04-02
ci(publish): fix publish job

f2prateek:
This adds filters to jobs that publish depends on so that the publish job can run on tags.
Reference: https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag.


f2prateek:
looks like it worked! https://circleci.com/workflow-run/491ac45e-502c-4a75-a11a-db214c92aba8

PR #55, opened 2019-04-04
feat(anonymousid): add getAnonymousId function

f2prateek:
This adds a `getAnonymousId` method to retrieve the current users anonymousId.

It uses `[SEGAnalytics.sharedAnalytics getAnonymousId]` on iOS and `analytics.getAnalyticsContext().traits().anonymousId()` on Android.

Usage:

```
const logAnonymousId = async () => {
	  console.log('anonymousId: %s', await analytics.getAnonymousId())
}
```

Also adds an example to the seed app and updates docs.


f2prateek:
Should look into promisifying this as well, but would require some changes in the `run` method for the native wrappers.

f2prateek:
iOS:
Screen Shot 2019-04-04 at 6 25 46 PM


Android:
Screen Shot 2019-04-04 at 6 45 52 PM

PR #65, opened 2019-04-27
identify should accept null for memberId, to match native APIs

patbenatar:
In the documentation it says "If you don't have a userId but want to record traits, you should pass nil." but the type signature is enforcing no nulls.


misterpoloy:
I think, this might be related to this issue:
https://github.com/segmentio/analytics-react-native/issues/107

bsneed:
Fixed in PR #147

bsneed:
Looks like I screwed this up, my apologies.

PR #73, opened 2019-06-06
Move application lifecycle tracking in Android to the native wrapper

### LINKED ISSUE #36: Android lifecycle events not captured
We are using `@segment/analytics-react-native - ^0.0.1-beta.3` in our application. But the Lifecycle events are not being tracked automatically in the case of Android, though it works fine in iOS. We tried to simulate the same using a new RN project and facing the same issue with Android.

```
"dependencies": {
    "@segment/analytics-ios": "github:segmentio/analytics-ios#3.6.10",
    "@segment/analytics-react-native": "^0.0.1-beta.3",
    "react": "16.6.3",
    "react-native": "0.58.3",
    "react-native-gesture-handler": "^1.0.15"
  },
  "devDependencies": {
    "babel-core": "^7.0.0-bridge.0",
    "babel-jest": "24.0.0",
    "jest": "24.0.0",
    "metro-react-native-babel-preset": "0.51.1",
    "react-test-renderer": "16.6.3"
  }
```

Test Repo:

https://github.com/anandwahed/test-rn-segmentsdk

kadikraman:
Fixes #36

## Bug
Application lifecycle events (Application Installed, Opened, Updated) in Android are not being tracked.

## Cause
In the analytics-android package, lifecycle events being tracked in the callback of [`onActivityCreated`](https://github.com/segmentio/analytics-android/blob/master/analytics/src/main/java/com/segment/analytics/Analytics.java#L292). However, neither `onActivityCreated` or `onActivityStarted` are ever called when using this library via the native wrapper. The other lifecycle methods, e.g. `onActivityResumed`, `onActivityPaused`, `onActivityStopped` etc are called as expected, which makes this a React Native specific bug.

The most likely reason for this is that `MainActivity` is created to run the React Native app before the native Analytics client is initialised, so that `onActivityCreated` and `onActivityStarted` for the `MainActivity` would have been triggered before the Analytics client has subscribed to them.

## Solution
Move the lifecycle tracking code into the native wrapper.

The logic for lifecycle tracking is mirrored exactly as in `analytics-android`, and the `trackAppLifecycleEvents` flag will no longer be passed down to the native package, but handled in the wrapper.

## Testing
Tested locally using Android Studio and a Nexus emulator, running Android 8.
Confirmed that each of the three lifecycle events are correctly sent and can be seen in the Live Debugger (`https://app.segment.com/[name]/sources/android/debugger`)

Screenshot 2019-06-06 at 11 37 27

### DIFF
--- a/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
+++ b/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
@@ -114,7 +114,7 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
         }
 
         val builder = Analytics
-                .Builder(reactApplicationContext, options.getString("writeKey"))
+                .Builder(reactApplicationContext, writeKey)
                 .flushQueueSize(options.getInt("flushAt"))
 
         if(options.getBoolean("recordScreenViews")) {
--- a/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
+++ b/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
@@ -24,11 +24,14 @@
 
 package com.segment.analytics.reactnative.core
 
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
 import com.facebook.react.bridge.*
 import com.segment.analytics.Analytics
 import com.segment.analytics.Properties
 import com.segment.analytics.Traits
 import com.segment.analytics.ValueMap
+import com.segment.analytics.internal.Utils.getSegmentSharedPreferences
 import java.util.concurrent.TimeUnit
 
 class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaModule(context) {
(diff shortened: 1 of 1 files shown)


packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -39,11 +42,67 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
...
+        val editor = sharedPreferences.edit()
+        editor.putString(versionKey, currentVersion)
+        editor.putInt(buildKey, currentBuild)
+        editor.apply()
     }
 
     @ReactMethod
     fun setup(options: ReadableMap, promise: Promise) {
         val json = options.getString("json")
+        val writeKey = options.getString("writeKey")
```
f2prateek: Is this used?

f2prateek:
Merged https://github.com/segmentio/analytics-react-native/pull/74. Let's rebase this, and the build should pick up the fixed config.

packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
kadikraman: Yes - look at line 117 and 70. It's used as a storage key for shared preferences.
Good spot though, I should have used the `writeKey` variable on line 117, but I forgot to replace it. Will go now!

PR #81, opened 2019-06-12
Feature/integrations

kadikraman:
## Feature
Add the ability to pass in integration specific options

## API
Similarly to the native [iOS](https://segment.com/docs/sources/mobile/ios/#selecting-destinations) and [Android](https://segment.com/docs/sources/mobile/android/#selecting-destinations) implementations, the `integrationOptions` will be passed in as an object as a third argument to `track`, `identify` or `group`, e.g:

```js
analytics.track('Event Name', { foo: "bar" }, {
  integrations: { 
    Mixpanel: false,
    "Google Analytics": false
  }
});
```

All integrationOptions default to true unless configured otherwise. In the example above, the track event will not be sent to Mixpanel or Google Analytics, but will be sent to any other integrations if there are any more defined.

### Android
_(Pixel 2 API 28)_
1. Sent the following events (the non-tracking one first!)
```js
// 1
analytics.track('Android - Google Analytics = false',
  { time: new Date().getTime(), platform: Platform.OS },
  { integrations: { 'Google Analytics': false } }
);

// 2
analytics.track('Android - Google Analytics = true',
  { time: new Date().getTime(), platform: Platform.OS },
  { integrations: { 'Google Analytics': true } }
);
```
2. Live Debugger output

Screenshot 2019-06-13 at 11 23 49

3. Google Analytics
Screenshot 2019-06-13 at 11 25 59

### iOS
_iPhone X_
1. Sent the following events (the non-tracking one first!)
```js
// 1
analytics.track('iOS - Google Analytics = false',
  { time: new Date().getTime(), platform: Platform.OS },
  { integrations: { 'Google Analytics': false } }
);

// 2
analytics.track('iOS - Google Analytics = true',
  { time: new Date().getTime(), platform: Platform.OS },
  { integrations: { 'Google Analytics': true } }
);
```
2. Live Debugger output
Screenshot 2019-06-13 at 14 03 29

3. Google Analytics
Screenshot 2019-06-13 at 14 02 52

### Not bundled
In the following examples, Google Analytics is bundled in the example app, and Mixpanel is not.
The expected behaviour is that Google Analytics will always be `false` and Mixpanel will have whatever value (boolean or object) that is passed in.
```js
analytics.track('Set to true', {  platform: Platform.OS },
  { integrations: { Mixpanel: true, "Google Analytics": true } }
);
```
#### Android
Screenshot 2019-06-14 at 11 42 56

#### iOS
Screenshot 2019-06-14 at 12 13 29

```js
analytics.track('Set to false', {  platform: Platform.OS },
  { integrations: { Mixpanel: false, "Google Analytics": false } }
);
```
#### Android
Screenshot 2019-06-14 at 11 46 51

#### iOS
Screenshot 2019-06-14 at 12 19 27

```js
analytics.track('Additional parameters', {  platform: Platform.OS },
  { integrations: { Mixpanel: { foo: "bar" }, "Google Analytics": { foo: "bar" } }
);
```
#### Android
Screenshot 2019-06-14 at 11 59 45

#### iOS
Screenshot 2019-06-14 at 12 21 36

### DIFF
--- a/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
+++ b/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
@@ -29,10 +29,14 @@ import android.content.pm.PackageManager
 import com.facebook.react.bridge.*
 import com.segment.analytics.Analytics
 import com.segment.analytics.Properties
+import com.segment.analytics.Options
 import com.segment.analytics.Traits
 import com.segment.analytics.ValueMap
 import com.segment.analytics.internal.Utils.getSegmentSharedPreferences
 import java.util.concurrent.TimeUnit
+import com.facebook.react.bridge.ReadableMap
+
+
 
 class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaModule(context) {
     override fun getName() = "RNAnalytics"
@@ -42,7 +46,6 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
 
     companion object {
         private var singletonJsonConfig: String? = null
-        private var key: String? = null
         private var versionKey = "version"
         private var buildKey = "build"
     }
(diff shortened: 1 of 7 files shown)


packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -56,18 +59,32 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
         }
     }
 
+    private fun getOptions(properties: ReadableMap?): Options {
```
f2prateek: From the invocation of this function, we pass in `options` as an argument, so should we rename this argument?

It might also be good to document this function, as it's not clear.

packages/core/src/__tests__/analytics.spec.ts
```diff
@@ -103,3 +103,31 @@ function testCall<K extends keyof typeof Bridge>(name: K) {
...
 	}) as (typeof Bridge)[K]
 }
+
+it('enables setting integrations from the middleware', async () => {
+	const integrations = {
+		'Google Analytics': false,
+		Mixpanel: { foo: 'bar' }
+	}
+
+	analytics.middleware(async ({ next, context, data, type }) => {
```
fathyb: We can use type-guard to smart-cast the payload:

```js
 	analytics.middleware(async event => {
		switch (event.type) {
			case 'track':
			case 'identify':
			case 'group':
				// event.type is 'track' | 'identify' | 'group'
				return next(context, { ...event.data, integrations })
			default:
				return next(context)
		}
```

packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -56,18 +59,41 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
...
+                when (integrations.getType(nextKey)) {
+                    ReadableType.Boolean -> options.setIntegration(nextKey, integrations.getBoolean(nextKey))
+                    ReadableType.Map -> {
+                        options.setIntegration(nextKey, true)
+                        options.setIntegrationOptions(nextKey, integrations.getMap(nextKey)?.toHashMap())
+                    }
+                }
+            }
+        }
+        return options;
```
fathyb: Extra semicolon?

PR #82, opened 2019-06-13
feat(ios): set trackAdvertising=true by default

kadikraman:
Update the React Native iOS configuration to enable tracking by default, same as the [native iOS library](https://github.com/segmentio/analytics-ios/blob/d1db92dab26b75ae7a8cba925604d41c464b0051/Analytics/Classes/SEGAnalyticsConfiguration.m#L55).


jeveloper:
Could you @kadikraman  tell us if this means when submitting the app/app update that IDFA is on? 
App Store and google play

thanks

Issue #30, opened 2018-12-19
RNAnalytics not found


migue1s:
I am consistently getting a `@segment/analytics-react-native-mixpanel/ios/main.m:10:9: 'RNAnalytics/RNAnalytics.h' file not found`when adding the mixpanel integration. I am currently integratitng the analytics-react-native library without cocoapods.

I have followed the steps in the readme and #17, running the project without the mixpanel integration will correctly send data to segment but not to mixpanel.

Jonovono:
@migue1s did you solve this? Having the same issue

migue1s:
I did not manage to solve it @Jonovono. The project I'm working on had some previous setup in segment to forward data to mixpanel and it worked (partially) without the integration.

caroaguilar:
Look at https://github.com/segmentio/analytics-react-native/issues/17 maybe that works for you

SnidelyWhiplash:
Same with Firebase, stuck all day on this.

I am not using CocoaPods because everything breaks with just about every RN lib I have when trying to do so. 🙄

This line from the README worries me a bit:
>We do not support sending data to bundled, device-mode integrations outside of Cocoapods.

Does that mean it's _impossible_ for me to use Firebase without somehow figuring out how to integrate Pods into my project? Is that what's causing this error, or..?

powersjcb:
+1 for Amplitude, tried a bunch of the suggested fixes, nothing seems to be working.

powersjcb:
I just uncovered more specific behavior about this failure.

- Building with both the base package and the plugin: Failure - RNAnalytics not found
- Building with just the base package: Success
- Building with the plugin after having first compiled the base package: Success.

I don't know the inner workings of how iOS apps are compiled, but I hope this helps someone figure out a workaround fix for this.

CapitanRedBeard:
Seems no one can get any integrations to work. 

Personally I'm trying to get the AppsFlyer integration working. I've gotten the core RNAnalytics project integrated and I pull down the Analytics SDK via Carthage and it works great. However I get this error after installing the integration and linking it

![image](https://user-images.githubusercontent.com/5445860/58134656-56281600-7bdc-11e9-8f0b-bdca70e84558.png)

I've tried adding the RNAnalytics search path to my Library search paths and also just adding it into the RNAnalyticsIntegration_* project. However it fails every time. 


[...]

PR #116, opened 2019-10-09
Fix for RNAnalytics not found issue (#30) 

emadgit:
**Description** 
`RNAnalytics` dependency is missing for all integrations, which needed when it imported in `ios/main.m` file `#import <RNAnalytics/RNAnalytics.h>`

Also the integration generator will generate factory_header with `Segment-` format which causing problem since the right format should be `Segment_` as an example : 
Current import ( Invalid ) : 
`#import <Segment-Adjust/SEGAdjustIntegrationFactory.h>`

Correct format : 
`#import <Segment_Adjust/SEGAdjustIntegrationFactory.h>`

Fix for issue below : 
https://github.com/segmentio/analytics-react-native/issues/30

### DIFF
--- a/packages/core/docs/README.md
+++ b/packages/core/docs/README.md
@@ -24,7 +24,7 @@
 
 **Ƭ Integration**: *`function` \| `object`*
 
-*Defined in [analytics.ts:8](https://github.com/emadgit/analytics-react-native/blob/master/packages/core/src/analytics.ts#L8)*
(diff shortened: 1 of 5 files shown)


prashantShrivastava91:
I am facing the same error while integrating with firebase. Is this PR going to me merged anytime soon.

snamoah:
will this be merged anytime soon? it's causing build fails for `@segmentio/analytics-react-native-firebase` also

KolenkoJan:
Can this be merged?

llaine:
Guys any estimation on when this is going to be merged?

How are you able to makes these changes in the node_modules, can you explain us @squirly ?

PR #140, opened 2020-03-10
make user id nullable in line with java interface

aultimus:
internal issue https://segment.atlassian.net/browse/LIB-1639

function definition disagreed with java on nullability of userId https://github.com/segmentio/analytics-android/blob/master/analytics/src/main/java/com/segment/analytics/Analytics.java#L440


bsneed:
Can you verify that all the other nullability settings in this file are correct?

bsneed:
This is addressed in PR #147

PR #158, opened 2020-04-08
Fix for build error, defaultKotlinVersion is still missing

emadgit:
`defaultKotlinVersion` is still missing in build.gradle in dependencies scope, when we run the android app we getting the below error : 

`Could not get unknown property 'defaultKotlinVersion' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
`

This PR will fix it.


emadgit:
A better [fix](https://github.com/segmentio/analytics-react-native/pull/159) is merged, so closing this one

PR #166, opened 2020-04-23
fix(android): add trackAttributionInformation method

snamoah:


bsneed:
This looks like a great addition @snamoah !  Looking forward to it landing!  👍

snamoah:
@bsneed thanks. quick question,  I have an issue after implementing this. [This line](https://github.com/segmentio/analytics-react-native/pull/166/files#diff-e77c1c9286a4ce094951a25cbe697b39R145) fails and after debugging, the error I get is

```
java.io.FileNotFoundException: https://mobile-service.segment.com/v1/attribution
```
I tried to mirror the same code from analytics-android [here](https://github.com/segmentio/analytics-android/blob/2a96f19afd5a7b6280af7d8c8a1cb7aa6ff5c69e/analytics/src/main/java/com/segment/analytics/Analytics.java#L302) I'm not sure if I'm missing something?

bsneed:
Hi @snamoah, you probably want to dig in and see what the actual output is.  That would come up if the server is rejecting your request for some reason.  It might not like the user agent or a number of other things, but you'd need to see the actual output of the response.

snamoah:
closing this pr as this has been fixed my your team 🙏

PR #168, opened 2020-04-23
fix(bridge): fix p(r)omise typo

nutlike:
PR #162 introduced a typo which fails debug builds, this PR fixes it.


bsneed:
Something is up with Snyk, so the test is failing.  Will come back to this at a later time.

bsneed:
Thanks @nutlike !!

PR #178, opened 2020-05-21
[Feature] add support for supplying defaultProjectSettings

prayansh:
## Summary
This change exposes the new setup configuration property to include default project settings in case Segment.com cannot be reached (preventing fetching project settings from Segment).

### DIFF
--- a/packages/integrations/integrations.yml
+++ b/packages/integrations/integrations.yml
@@ -26,7 +26,7 @@
 - name: Intercom
   ios:
     pod:
-      version: 1.0.0-beta
+      version: 1.0.1-beta
 
 - name: Tapstream
   ios:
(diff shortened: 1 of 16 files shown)


packages/integrations/integrations.yml
```diff
@@ -26,7 +26,7 @@
 - name: Intercom
   ios:
     pod:
-      version: 1.0.0-beta
```
prayansh, line 29: do we need to upgrade these now that we are still on 3.+ series?

bsneed: No, since I back ported the iOS bits to 3.9, no changes necessary to the integrations.

packages/integrations/integrations.yml
```diff
@@ -57,6 +57,9 @@
       repo: https://comscore.bintray.com/Analytics
 
 - name: Taplytics
+  ios:
+    pod:
+      version: 1.1.3
```
prayansh, line 62: we dont need these upgrades either right?

prayansh: @bsneed

PR #181, opened 2020-06-05
fix(android): add missing @

snamoah:
This is an an issue causing android builds to fail. See attached screenshot. This is after updating to the 1.1.1-beta.5

Screenshot 2020-06-05 at 15 38 21


bsneed:
You need to upgrade to RN 0.62 to fix this.  We can't support multiple versions of react-native effectively.

PR #184, opened 2020-06-12
feat(android): allow passing integration options

squirly:
Add support for passing object integration data on Android.

### DIFF
--- a/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
+++ b/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
@@ -245,7 +245,11 @@ private fun optionsFrom(context: ReadableMap?, integrations: ReadableMap?): Opti
     }
 
     integrations?.toHashMap()?.forEach { (key, value) ->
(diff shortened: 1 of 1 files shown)


packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -245,7 +245,11 @@ private fun optionsFrom(context: ReadableMap?, integrations: ReadableMap?): Opti
     }
 
     integrations?.toHashMap()?.forEach { (key, value) ->
-        options.setIntegration(key, value.toString().toBoolean())
+        if (value is HashMap<*, *>) {
+            options.setIntegrationOptions(key, value.toMap() as Map<String, Object>)
```
prayansh, line 249: @squirly could you change this to use `Any` rather than `Object` since its more appropriate for Kotlin

prayansh:
```suggestion
            options.setIntegrationOptions(key, value.toMap() as Map<String, Any>)
```

PR #192, opened 2020-07-08
Improve installation section of README

kelset:
Given the requirement in the readme of RN 0.62, it's better to replace the link command with the pod install step, as required by the [autolinking cli documentation](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md).

(in fact, the link command doesn't do the pod install so the installation would be incomplete without it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ If you are part of a new startup  (&lt;$5M raised, &lt;2 years since founding), 
 
 ```bash
 $ yarn add @segment/analytics-react-native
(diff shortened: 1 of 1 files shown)


bsneed:
thanks so much @kelset !

teekirol:
Should this also apply for the device-mode destination packages (which also instruct to run the link command)?

kelset:
yeah actually it should, I didn't check the whole readme when submitting the fix 😅

PR #202, opened 2020-08-03
feat(core): http proxying via setup configuration

adkenyon:
extended the configuration API to allow directing analytics traffic through a hosted proxy (or other)

I noticed that the native [iOS](https://segment.com/docs/connections/sources/catalog/libraries/mobile/ios/#proxy-http-calls) and [Android](https://segment.com/docs/connections/sources/catalog/libraries/mobile/android/#proxy-http-calls) SDK's had support for proxying http requests. The current RN module did not have this capability so I simply ported it over.


erheme318:
@bsneed, @prayansh  Please Review

PR #206, opened 2020-08-11
fix(ios): fixes for building for newer rn

bsneed:
- Fixes issues w/ header slug creation.
- Fixes issues w/ non-segment owned integrations not building and needing updates.
- Fixes issue w/ running test app w/ Facebook-app-events integration.
- Fixes pod spec mismatch between analytics-ios and analytics-react-native.
- Fixes issue w/ package.json not referencing correction versions for supported RN builds.
- Fixes pod spec to use new format of imports for RN 0.62+

### DIFF
--- a/packages/core/RNAnalytics.podspec
+++ b/packages/core/RNAnalytics.podspec
@@ -16,11 +16,11 @@ Pod::Spec.new do |s|
   s.author              = { 'Segment' => 'friends@segment.com' }
   s.source              = { :git => 'https://github.com/segmentio/analytics-react-native.git', :tag => s.version.to_s }
 
-  s.platform            = :ios, '10.0'
+  s.platform            = :ios, '11.0'
   s.source_files        = 'ios/**/*.{m,h}'
   s.static_framework    = true
 
-  s.dependency          'Analytics'
+  s.dependency          'Analytics', '~> 4.0.4'
   s.dependency          'React'
 end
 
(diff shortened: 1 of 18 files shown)


packages/core/RNAnalytics.podspec
```diff
@@ -16,11 +16,11 @@ Pod::Spec.new do |s|
   s.author              = { 'Segment' => 'friends@segment.com' }
   s.source              = { :git => 'https://github.com/segmentio/analytics-react-native.git', :tag => s.version.to_s }
 
-  s.platform            = :ios, '10.0'
+  s.platform            = :ios, '11.0'
```
oomathias, line 19: @bsneed Is this required? It's a breaking change.

https://github.com/facebook/react-native#-requirements still supports `10.0`

bsneed: Yes, it is required.  Our iOS lib minimum is now 11.0.  That's my bad on not bumping the version #'s appropriately given it was indeed a breaking change.  My apologies for that.  If you need 10.x support, we recommend you move back to a previous version.

oomathias: @bsneed Ok thanks but unfortunatly I can't because we upgraded to the latest react native version and need to support iOS 10.
I will try to see if it works for us by patching this file back to iOS 10 (at least it builds, but I need to do regression testing).

PR #213, opened 2020-09-08
Put minimum deployment target to iOS 10.0

reinismu:
Revert target changes made in https://github.com/segmentio/analytics-react-native/pull/206

If there is a reason for it to be iOS 11.0 please let us know as there is nothing about it in PR. As mentioned https://github.com/segmentio/analytics-react-native/pull/206#discussion_r481255148 here React Native still supports iOS 10.0

I did a test with iOS 10.3.1 and everything seemed to work fine.

### DIFF
--- a/packages/core/RNAnalytics.podspec
+++ b/packages/core/RNAnalytics.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
   s.author              = { 'Segment' => 'friends@segment.com' }
   s.source              = { :git => 'https://github.com/segmentio/analytics-react-native.git', :tag => s.version.to_s }
 
-  s.platform            = :ios, '11.0'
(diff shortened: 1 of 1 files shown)


bsneed:
Yes, it is required. Our iOS lib minimum is now 11.0. That's my bad on not bumping the version #'s appropriately given it was indeed a breaking change. My apologies for that. If you need 10.x support, we recommend you move back to a previous version.

reinismu:
> Yes, it is required. Our iOS lib minimum is now 11.0. That's my bad on not bumping the version #'s appropriately given it was indeed a breaking change. My apologies for that. If you need 10.x support, we recommend you move back to a previous version.

Thanks for reply :) Out of curiosity why is 11.0 required minimum version now?

Issue #204, opened 2020-08-04
Adding undefined to JsonValue type


pstakoun:
The `JsonValue` type in `bridge.ts` doesn't include `undefined`. The specific use case for me is tracking optional parameters like `obj?.optional_param` rather than having to use additional nullish coalescing logic. Would we be open to changing:
```
export type JsonValue = boolean | number | string | null | JsonList | JsonMap
``` 
to
```
export type JsonValue = boolean | number | string | null | undefined | JsonList | JsonMap
```

PR #221, opened 2020-09-22
add undefined to JsonValue type union

mngyuan:
`JsonValue` should allow `undefined` for the optional chaining operator `?.` see https://github.com/segmentio/analytics-react-native/issues/204
AFAIK this should be typesafe because we're strictly increasing the leniency of the `JsonValue` type but curious what thoughts are on this

### DIFF
--- a/packages/core/src/bridge.ts
+++ b/packages/core/src/bridge.ts
@@ -26,7 +26,7 @@ export interface Configuration {
 	}
 }
 
-export type JsonValue = boolean | number | string | null | JsonList | JsonMap
+export type JsonValue = boolean | number | string | null | undefined | JsonList | JsonMap
 export interface JsonMap {
 	[key: string]: JsonValue
 	[index: number]: JsonValue


prayansh:
I am hesitant to add the `undefined` value bcos it is not a valid json value as stated here https://www.w3schools.com/js/js_json_datatypes.asp

But in any case shouldnt optional chaining work with `null`? For [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining): 
```
The ?. operator functions similarly to the . chaining operator, except that instead of causing an error if a reference is nullish (null or undefined), the expression short-circuits with a return value of undefined.
```
So you could do something like `let x: JsonValue = obj?.missingProp ?? null`

mngyuan:
ah ok, thanks for the clarification, I was not aware `undefined` was not in the JSON spec. it is possible to use the `??` to return a `null` in case of `undefined`, but I find this to be not terribly readable, that is, when reading 

```
let x: JsonValue = obj?.missingProp ?? null
```

my initial response is to say that `?? null` seems to be a no-op, especially since the name 'null coalescing operator' visibly references `null`, not `undefined`, in its name, and it isn't immediately clear what coalescing means in this context.

i suppose this is something I should take up with ECMAScript / JSON spec folks though.

PR #235, opened 2020-11-01
chore: adding JsonMap type to Context

dbilgin:
As the context object has the library inside and the user has to define it when they are defining the context, it causes Typescript errors when the user doesn't set the library object.
Setting the IP directly without the library still results in the tracking below:
Screenshot 2020-11-01 at 23 57 35
So I thought having a generic context type with a suggestion of library inside would be good instead of the current version.

### DIFF
--- a/packages/core/docs/README.md
+++ b/packages/core/docs/README.md
@@ -24,7 +24,7 @@
 
 **Ƭ Integration**: *`function` \| `object`*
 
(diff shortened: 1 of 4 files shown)


alanjcharles:
@dbilgin added this in #297 to get passing tests. It has been merged and will be released later this week.

dbilgin:
@alanjcharles I don't get why this PR wouldn't be merged instead, but thanks for the update

PR #253, opened 2021-01-08
Fix crash when using proxy on iOS 12

leethree:
`proxyOptions[@"scheme"]` is already a string. Calling `stringValue` on `NSString` causes the following runtime crash on iOS <=12. 

```
NSInvalidArgumentException
-[NSTaggedPointerString stringValue]: unrecognized selector sent to instance 0xa000073707474685
```

Debugger says it's invalid:

image


bsneed (approved):

bsneed:
great catch @leethree, thank you!!

PR #259, opened 2021-01-28
Fix support for setting custom anonymous IDs from React Native library.

carlos-kelly:


packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -228,12 +228,24 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
...
+    fun identify(userId: String?, traits: ReadableMap?, options: ReadableMap?, integrations: ReadableMap?, context: ReadableMap?) {
+
+        val mergedTraits = if (options?.hasKey("anonymousId") == true) {
+            val map = WritableNativeMap()
+            map.merge(traits)
+            map.putString("anonymousId", options.getString("anonymousId"))
+            map
+        } else {
+            traits
+        }
```
carlos-kelly, line 240: The iOS library expects the `anonymousId` to be part of the `options` object where the Android library looks for it at creation time inside `traits`.

PR #282, opened 2021-03-15
Update Pod.podspec

paulxuca:
Fixes issues in later versions of React Native


alanjcharles:
Hi @paulxuca thanks for bringing this to our attention! Do you have any additional information as to when this was released and which versions of RN it will be an issue for? Thanks!

alanjcharles:
updated in the latest release https://github.com/segmentio/analytics-react-native/blob/master/packages/integrations/template/Pod.podspec

Issue #255, opened 2021-01-24
[Android] Error on setup: "Method addObserver must be called on the main thread"


cristeahub:
Hi,

After upgrading one of my other dependencies in my android app (namely Stripe Android SDK from 10.x to 16.x) I surprisingly encountered an error in this SDK. Note that one of the changes from this huge jump is that the StripeSDK has migrated to AndroidX.

When i do `analytics.setup(MY_WRITE_KEY, {some: 'options'}` I get a redbox in my app with the text `writeKey must not be null or empty`.

I have debugged and verified that the writeKey is indeed set correctly, so this is just a symptom and not the actual issue.

Digging further down I found that when initializing the Segment Android SDK I got the following error `Method addObserver must be called on the main thread`. This seems to be the actual issue causing my subsequent calls to `analytics` to fail due to the writeKey not being set.

The error is caught here:

https://github.com/segmentio/analytics-react-native/blob/226a9e4a5010b116d173da80ca1d335a5b7b01b9/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt#L187-L191

This error comes from the `analytics-android` package that this package depends on. The error seems to stem from AndroidX issues, and there's even an issue on that repo for the error: https://github.com/segmentio/analytics-android/issues/725 .

This is a blocker for us now, I am asking here since we are using react native and there was no answer from the developers at the other repo. Is there a way to escalate this issue so it gets attention and gets fixed?

I found a related issue from a different vendor here: https://github.com/getsentry/sentry-android/issues/520 . This issue was about lifetime handlers, which both the original issue creator on the segment-android repo and myself use. However simply turning that off was not enough to fix the issue.

Alternatively do you know of any workarounds?

Thanks for a great library otherwise!

wmonecke:
@cristeahub Hi there! I am currently facing the same issue with another RN library. 
Did you manage to solve your issue? :)

cristeahub:
> @cristeahub Hi there! I am currently facing the same issue with another RN library.
> Did you manage to solve your issue? :)

There's a chain of issues here, but https://github.com/segmentio/analytics-react-native/pull/256 fixed the issue for me.

You can read more on this issue: https://github.com/segmentio/analytics-android/issues/725 and I used this patch here at one point to fix it: https://github.com/segmentio/analytics-android/issues/725#issuecomment-767069769 .

wmonecke:
[...]

PR #285, opened 2021-03-18
fix(android): lifecycle option not forwarded

sterlingwes:
I'm hitting the issue in https://github.com/segmentio/analytics-react-native/issues/255 with the latest release, and followed the instructions in https://github.com/segmentio/analytics-react-native/pull/256 to attempt to resolve this but found out that the required option (`experimentalUseNewLifecycleMethods`) is not making its way to the native module `setup()` method.

It's unclear to me why there's typings in both the Bridge and the Analytics js modules, perhaps that's where the confusion lies?

### DIFF
--- a/packages/core/src/bridge.ts
+++ b/packages/core/src/bridge.ts
@@ -18,6 +18,7 @@ export interface Configuration {
 	android: {
 		flushInterval?: number
 		collectDeviceId: boolean
+		experimentalUseNewLifecycleMethods: boolean
 	}
 	ios: {
 		trackAdvertising: boolean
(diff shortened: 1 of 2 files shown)


antoinefBam:
Hello 😇

I have the same problem as the author.

I am not currently able to build my app in release mode containing both [Segment](https://github.com/segmentio/analytics-react-native) and [RevenueCat](https://github.com/RevenueCat/react-native-purchases) SDKs. The only fix I found is to disable `experimentalUseNewLifecycleMethods` inside `Segment` setup. But as the author the option is not correctly forwarded to the native module so I have to manually disable it in `node_modules`... 😬

Is there any chance that this PR will be merged soon?

alanjcharles:
Hi, thanks for bringing this to our attention. We have merged a fix for this here https://github.com/segmentio/analytics-react-native/pull/298 and are planning a release here in the next few days.

PR #287, opened 2021-03-22
Don't use GA, use firebase

sanscontext:
Remove GA from the list of react device-mode sdks. Google forced folks to move mobile implementations to Firebase.


alanjcharles (approved):
Looks good to me

PR #298, opened 2021-04-16
Bump Analytics-android dependency

prayansh:
- Bump analytics-android to 4.9.3
- Fix nullable write-key issue
- Disable experimental lifecycle methods

### DIFF
--- a/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
+++ b/packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
@@ -106,6 +106,10 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
         val json = options.getString("json")
         val writeKey = options.getString("writeKey")
 
+        if (writeKey == null) {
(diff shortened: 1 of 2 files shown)


alanjcharles (approved):
looks good to me

packages/core/android/src/main/java/com/segment/analytics/reactnative/core/RNAnalyticsModule.kt
```diff
@@ -196,14 +207,6 @@ class RNAnalyticsModule(context: ReactApplicationContext): ReactContextBaseJavaM
             return promise.reject("E_SEGMENT_ERROR", e)
         }
 
-        if(options.getBoolean("trackAppLifecycleEvents")) {
-            this.trackApplicationLifecycleEvents(writeKey)
```
nfcampos, line 200: Could you explain this change?
After removing this line the private method defined in line 65 isn't called anywhere, sounds like a bug?

prayansh: yup this was a bug. @nfcampos thanks for pointing it out :), It really helped us solve a crucial bug 🙌

Issue #323, opened 2021-05-19
 The logging tag can be at most 23 characters, was 31


copilots-app:
I'm getting this when building RN 0.64.1:

integration/firebase/IntegrationModule.kt:43: Error: The logging tag can be at most 23 characters, was 31 (RNAnalyticsIntegration_Firebase) [LongLogTag]
              Log.v("RNAnalyticsIntegration_Firebase", "Firebase integration ready.")
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                    
Is it fixable?

alanjcharles:
@copilots-app thanks for letting us know! Merged a fix here: https://github.com/segmentio/analytics-react-native/pull/325

PR #325, opened 2021-05-19
fix: change logging variable to constant string resolves error on 0.64.1

alanjcharles:
addresses issue: https://github.com/segmentio/analytics-react-native/issues/323


packages/integrations/template/android/src/main/java/com/segment/analytics/reactnative/integration/Module.kt
```diff
@@ -40,7 +40,7 @@ class {{{nativeModule}}}Module(context: ReactApplicationContext): ReactContextBa
         RNAnalytics.addIntegration({{{factoryClass}}}.FACTORY)
 
         RNAnalytics.addOnReadyCallback("{{{slug}}}", Analytics.Callback { instance ->
-            Log.v("{{{nativeModule}}}", "{{{slug}}} integration ready.")
+            Log.v("Native Module", "{{{slug}}} integration ready.")
```
prayansh, line 43: lets use `RNAnalyticsIntegration`

prayansh (approved):
just change the string and we good to go

PR #328, opened 2021-05-26
Fix Tracking Application Installed and Application Opened on cold-boot

prayansh:
This change re-adds the invocation of the function to track `Application Installed` and `Application Opened` events on cold-boot


alanjcharles (approved):
LGTM

PR #329, opened 2021-05-28
fix: correct import of Segment_Facebook_App_Events

BerniWittmann:
we had the problem, that the import file import <Segment-FacebookAppEvents/SEGFacebookAppEventsIntegrationFactory.h> could not be found, which caused our build to break. After digging around we found the solution that manually adding `Segment_Facebook_App_Events.framework` to the linked Libraries in xcode and adjusting the import fixes this.


alanjcharles:
@BerniWittmann I am able to build without an issue on `analytics-react-native 1.4.8` and `analytics-react-native-facebook-app-events-ios 1.4.8` this update won't work as expected and since this PR isn't passing our CI, it's not something we can merge. Please make sure you're on the latest version for both and if you're still running into issues, feel free to reach out to friends@segment.com. They can help file and prioritize a bug if the issue persists.

BerniWittmann:
@alanjcharles If you take a look at the CI logs, you will see that the failure is not related to this change and due to snyk not correctly initializing: `Set 'context: snyk' in your .circleci/config.yml workflow`

PR #371, opened 2021-10-29
Release it

alanjcharles:
yarn `yarn release-it --dry-run` on a clean master (it will error out otherwise)  to see it publish all packages. Let me know if there's more customization we should do. This readme is really good 

https://github.com/rwjblue/release-it-yarn-workspaces


oscb (approved):
Looks good!

PR #374, opened 2021-11-01
Fixup `yarn update` -> `yarn upgrade`

jgable:
Docs for `yarn upgrade` here; https://classic.yarnpkg.com/lang/en/docs/cli/upgrade/


alanjcharles (approved):
ah thank you! I always mix those up.

PR #377, opened 2021-11-03
Fb app events

alanjcharles:
rough initial commit to share progress/ make sure i accounted for Oscar's changes.


packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+  },
+};
+
+const mappedPropNames = generateMapTransform(mapEventProps, transformMap);
+
+const sanitizeName = (name: string) => {
+  //Facebook expects '_' instead of '.'
+  // and only accepts 40 characters in Event Names
+  let re = /\./gi;
+  let sanitizedName = name.replace(re, '_');
```
oscb, line 46: I think you can just do `name.replace('.', '_')`

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+): { [key: string]: string | number } => {
+  let productCount = 0;
+  let products = event.properties.products || [];
+  let params = {};
+
+  Object.keys(
+    event.properties.forEach((property: string) => {
+      if (property === 'fb_num_items') {
+        productCount = event.properties.fb_num_items;
+      } else if (property in mapEventNames) {
```
oscb, line 61: Should be `property in Object.values(mapEventProps)` so that it checks the transformed event properties keys to prioritize those (right now it checks the values part of mapEventNames

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+      }
+    })
+  );
+
+  if (products.length) {
+    products.forEach((product: any) => {
+      productCount++;
+      return product;
+    });
+  }
```
oscb, line 75: `productCount = products.length` instead ?

oscb: actually you can do this:
`const productCount = event.properties.fb_num_items ?? products.length`

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+  };
+};
+
+export default async (event: TrackEventType) => {
+  const safeEvent = mappedPropNames(event);
+  let convertedName = safeEvent.event as string;
+  let safeName = sanitizeName(convertedName);
+  let safeProps = fbParams(safeEvent);
+  let currency = 'USD';
+  if (Object.values(safeProps).includes('fb_currency')) {
```
oscb, line 91: Shouldn't it be `Object.keys(safeProps)` ? 
Also you can do this which might be more readable:
`if (safeProps.fb_currency !== undefined)`

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+    currency = safeProps.fb_currency as string;
+  }
+  if (Object.values(safeProps).includes('_valueToSum')) {
```
oscb, line 94: Same as above. Also it looks like currency is only important if you do `logPurchase` so you can move all that inside this if.

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+const sanitizeName = (name: string) => {
+  //Facebook expects '_' instead of '.'
+  // and only accepts 40 characters in Event Names
+  let re = /\./gi;
+  let sanitizedName = name.replace(re, '_');
+  return sanitizedName.substring(0, 40);
+};
+
+const fbParams = (
+  event: Record<string, any>
```
oscb, line 51: I would suggest not passing the full event, just the `properties` object. The input arg type would be the same but you wouldn't have to do `event.properties`. 
On the caller you just assign it back to the properties field: f.e.
`event.properties = fbParams(event.properties)`

That way we can leverage this if there's any other place in the future that also required a similar transformation without having it coupled to the `event.properties` structure that FB uses right now.

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const sanitizeName = (name: string) => {
+  //Facebook expects '_' instead of '.'
+  // and only accepts 40 characters in Event Names
+  let re = /\./gi;
+  let sanitizedName = name.replace(re, '_');
+  return sanitizedName.substring(0, 40);
+};
+
+const fbParams = (
```
oscb, line 50: Maybe rename to `sanitizeParams` ?

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+  event: Record<string, any>
+): { [key: string]: string | number } => {
+  let productCount = 0;
+  let products = event.properties.products || [];
```
oscb, line 54: I prefer the `??` to `||` to avoid truthy/falsy for type safety

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (products.length) {
+    products.forEach((product: any) => {
+      productCount++;
+      return product;
+    });
+  }
+
+  return {
+    ...params,
+    fb_num_items: productCount,
```
oscb, line 79: should we send `fb_num_items` as 0 when no products exist? maybe we can keep it undefined in that scenario instead

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const mappedPropNames = generateMapTransform(mapEventProps, transformMap);
+
+const sanitizeName = (name: string) => {
+  //Facebook expects '_' instead of '.'
+  // and only accepts 40 characters in Event Names
+  let re = /\./gi;
+  let sanitizedName = name.replace(re, '_');
+  return sanitizedName.substring(0, 40);
```
oscb, line 47: Let's put 40 in a const outside the class:
`const MAX_CHARACTERS_EVENT_NAME = 40;`

packages/plugins/plugin-appsflyer/README.md
```diff
@@ -9,13 +9,13 @@ You can provide [additional options](https://github.com/AppsFlyerSDK/appsflyer-r
 
 import { createClient } from '@segment/analytics-react-native';
 
-import { AppsflyerPlugin } from '@segment/analytics-react-native-plugin-appsflyer';
+import { FacebookAppEventsPlugin } from '@segment/analytics-react-native-plugin-appsflyer';
```
oscb, line 12: I think this is the incorrect file for this change he he

example/src/App.tsx
```diff
@@ -22,18 +22,22 @@ import { Logger } from './plugins/Logger';
 import { IdfaPlugin } from '@segment/analytics-react-native-plugin-idfa';
 // @ts-ignore
 import { AmplitudeSessionPlugin } from '@segment/analytics-react-native-plugin-amplitude-session';
+// @ts-ignore
+import { FacebookAppEventsPlugin } from '@segment/analytics-react-native-plugin-facebook-app-events';
 
 const segmentClient = createClient({
-  writeKey: 'WRITE KEY',
+  writeKey: 'BRoigAmFuQv7x5YhIWlvHJeKzqT6Yn5g',
```
oscb, line 29: Unintentional, probably

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
oscb: you can also do the currency thing in one line:
`const currency = (safe.Props.fb_currency as string | undefined) ?? 'USD';

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
alanjcharles: i think I can remove this conditional completely once adding ` const productCount = event.properties.fb_num_items ?? products.length`

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
alanjcharles: lol yup that's what it was in my head

example/src/App.tsx
alanjcharles: sure is lol

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,93 @@
...
+const fbParams = (
+  event: Record<string, any>
+): { [key: string]: string | number } => {
+  let products = event.properties.products ?? [];
+  const productCount =
+    (event.properties.fb_num_items || products.length) ?? undefined;
+  let params = {};
+
+  Object.keys(
+    event.properties.forEach((property: string) => {
```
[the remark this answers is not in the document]
oscb: (As it is right now it is actually iterating over the values and then getting the keys of the object)

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (property in mapEventNames) {
```
oscb, line 62:
```suggestion
      if (property in Object.values(mapEventProps)) {
```
So that it checks the property names, `Object.values` is important too so that it checks against the FB prop names (the transform resulting ones) else it checks the keys.

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,93 @@
...
+        params = {
+          ...params,
+          [property]: event.properties[property],
+        };
```
oscb, line 66: nit: this actually works but destructuring is not the most efficient operation in JS. Given that params is already a new dictionary you could do
```suggestion
        params[property] = event.properties[property];
```

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
oscb: Bringing this back, I think it is important for maintainability

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
```diff
@@ -0,0 +1,93 @@
...
+const transformMap: { [key: string]: (value: any) => any } = {
+  event: (value: string) => {
+    if (value in mapEventNames) {
+      return mapEventNames[value];
+    }
+    return value;
+  },
+};
+
+const mappedPropNames = generateMapTransform(mapEventProps, transformMap);
```
oscb, line 42: Given that the output of `generateMapTransform` is a function how about renaming it to something like `mapPropNames` or `sanitizeEvent` ?

oscb (changes requested):

packages/plugins/plugin-facebook-app-events/src/FacebookAppEventsPlugin.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Settings } from 'react-native-fbsdk-next';
+
+export class FacebookAppEventsPlugin extends DestinationPlugin {
+  type = PluginType.destination;
+  key = 'FacebookAppEvents';
+
+  async configure(analytics: SegmentClient) {
+    this.analytics = analytics;
+    let adTrackingEnabled =
+      this.analytics?.store.getState().main.context?.device?.adTrackingEnabled;
```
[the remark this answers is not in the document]
alanjcharles: yeah we also need a better way to update plugins if/when it becomes available because I'm not sure this is actually working correctly (will get to it when your comments are all resolved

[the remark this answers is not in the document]
alanjcharles: I wonder if there is a way to incorporate consent manager and idfa plugins easily into dest plugins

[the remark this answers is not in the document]
alanjcharles: like as an option

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
alanjcharles: yeah meant to catch that sorry about that

oscb: Reminder to change this name

packages/plugins/plugin-facebook-app-events/src/methods/screen.ts
```diff
@@ -0,0 +1,25 @@
+import type { ScreenEventType } from '@segment/analytics-react-native';
+import { AppEventsLogger } from 'react-native-fbsdk-next';
+
+// FB Event Names must be <= 40 characters
+// 'Viewed' and 'Screen' with spaces take up 14
+const MAX_CHARACTERS_EVENT_NAME = 26;
```
oscb, line 6: nit: How about putting it like this so that if we change it it is all automatically handled:
```suggestion
const PREFIX = 'Viewed ';
const SUFFIX = ' Screen';
const MAX_CHARACTERS_EVENT_NAME = 40 - PREFIX.length - SUFFIX.length;
```
Down below you can return the sanitizedName as:
```
return `${PREFIX}${trimmedName}${SUFFIX}`;
```

packages/plugins/plugin-facebook-app-events/src/methods/__tests__/track.test.ts
```diff
@@ -0,0 +1,83 @@
...
+jest.mock('react-native-fbsdk-next', () => ({
+  AppEventsLogger: {
+    logEvent: jest.fn(),
+    logPurchase: jest.fn(),
+  },
+}));
+
+// jest.mock('../../../../core/src/context', () => ({
+//   getContext: jest.fn().mockResolvedValue({}),
+// }));
```
oscb, line 14: Let's remove this!

packages/plugins/plugin-facebook-app-events/src/methods/screen.ts
```diff
@@ -0,0 +1,25 @@
...
+  return `${PREFIX} ${trimmedName} ${SUFFIX}`;
+};
+
+const sanitizeEvent = (
+  event: Record<string, any>
+): { [key: string]: string | number } => {
+  let properties = event.properties ?? {};
+
+  return { ...properties };
+};
```
oscb, line 19: Doesn't this need to do the same map transform as the track event with the properties?

packages/plugins/plugin-facebook-app-events/src/methods/track.ts
oscb: Rename to `mapPropNames` so that it looks more like a function

oscb (approved):

PR #378, opened 2021-11-06
Point to @react-native-async-storage/async-storage dependency

RubenGuerrero:
The dependency https://www.npmjs.com/package/@react-native-community/async-storage is deprecated if favour of https://www.npmjs.com/package/@react-native-async-storage/async-storage


oscb (approved):
This looks good! Thanks for taking care of this.

rafael-antunes-natoora:
Any ETA on when this will get to the released version? You guys say it works with expo, but since you are using the deprecated library, it's not compatible with expo...

alanjcharles:
@rafael-antunes-natoora we are planning to push a new release by the end of January. We have quite a few changes to account for so we are planning to roll everything into a major release and update to beta as soon as we can.

PR #379, opened 2021-11-09
fix: custom event integrations override support, refactor integration merge logic

oscb:


alanjcharles (approved):
approving since we walked through it together already but as always lgtm!

PR #382, opened 2021-11-17
chore: refactoring analytics client to lock down public interface

oscb:


alanjcharles (approved):
awesome man cannot wait to use these

PR #383, opened 2021-11-19
chore: upgrade react native to 0.66

oscb:
- Upgraded RN to the latest version
- Upgraded TS to 4.4.4 (Redux types in TS 4.5 have an issue right now)
- Upgraded several other libraries to new minor versions
- Consolidated repo commands to use `build` and `test` as standard, instead of `e2e:cmd` and so on. Changed GH actions to accommodate for the changes
- Moves the dependencies from the root to each individual package to avoid relying on "phantom dependencies"


alanjcharles (approved):
lgtm!

PR #384, opened 2021-11-23
fix: #identify argument type

raminos:
Hello library maintainers 👋 ,

thanks for this lovely library! @alan-eu we are happily using your service but we came across an incoherence between your docs and your code which I attempt to fix in this PR.

I read your contributing guidelines but couldn't find a pull request template. I hope the format I chose is okay.

## Problem

In your [docs](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#identify) you state for the `userId` argument of the `identify` method:

> The database ID for this user. If you don’t know who the user is yet, you can omit the `userId` and just record `traits`.

But with the current implementation Typescript won't allow us to omit a `userId`. We circumvent this issue by passing `undefined as any` but this is suboptimal.

## Solution

I tried to adjust the `identify` method's behaviour in a way that is in line with the existing logic, namely:

- `userId` can be `undefined`
- if `userId` is `undefined` the store is not updated

## Tests

I added a test case checking that only the user traits are updated when passing `undefined` as a `userId`. I tried to follow the preexisting unit test style.

As stated in the guidelines I ran all the scripts mentioned to check that tests, linting and the typescript compiler pass.

I'm looking forward to your comments!

Best,
Ramin

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -502,7 +502,7 @@ export class SegmentClient {
 
     this.store.userInfo.set({
       ...userInfo,
-      userId: userId,
+      ...(userId !== undefined ? { userId } : {}),
       traits: mergedTraits,
     });
 
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -487,7 +487,7 @@ export class SegmentClient {
     this.logger.info('TRACK event saved', event);
   }
 
-  identify(userId: string, userTraits?: UserTraits) {
+  identify(userId?: string, userTraits?: UserTraits) {
     const userInfo = this.store.userInfo.get();
     const { traits: currentUserTraits } = userInfo;
 
--- a/packages/core/src/__tests__/methods/identify.test.ts
+++ b/packages/core/src/__tests__/methods/identify.test.ts
@@ -79,4 +79,17 @@ describe('methods #identify', () => {
       userId: 'new-user-id',
     });
   });
+  it('does not update user id when user id is undefined', () => {
+    const client = new SegmentClient(defaultClientSettings);
+    jest.spyOn(client, 'process');
+
+    client.identify(undefined, { name: 'Mary' });
+
+    // @ts-ignore
+    expect(client.store.dispatch).toHaveBeenCalledTimes(1);
+    // @ts-ignore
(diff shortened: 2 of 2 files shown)


packages/core/src/analytics.ts
```diff
@@ -558,7 +558,7 @@ export class SegmentClient {
     this.logger.info('TRACK event saved', event);
   }
 
-  identify(userId: string, userTraits?: UserTraits) {
+  identify(userId: string | undefined, userTraits?: UserTraits) {
```
oscb, line 561: For consistency, lets use the optional as:
```suggestion
  identify(userId?: string, userTraits?: UserTraits) {
```

packages/core/src/analytics.ts
```diff
@@ -568,7 +568,9 @@ export class SegmentClient {
       },
     });
 
-    this.store.dispatch(this.actions.userInfo.setUserId({ userId }));
+    if (userId) {
```
oscb, line 571: Let's put a strict undefined check here:
```suggestion
    if (userId !== undefined) {
```

oscb (changes requested):
Thanks so much for noticing this and taking action! 
I left some minor suggestions but otherwise this looks good!

raminos:
Thanks @oscb for your fast and positive feedback! I applied your suggestions 😊

alanjcharles:
Hi @raminos Happy New Year! looks like there is still one comment left in our review. Happy to merge once that's been accounted for.

raminos:
> Hi @raminos Happy New Year! looks like there is still one comment left in our review. Happy to merge once that's been accounted for.

Hi @alanjcharles I totally missed that I didn't update the second comment! Thanks for reminding me 😍  commit is comming in a sec!

raminos:
@alanjcharles the code changed a bit in the meantime. The ternary operator is not super pretty but it gets the work done. Is there something I should change?

alanjcharles:
@raminos sorry for the delay!! we've had to take some unexpected time out of office in the past few weeks. Thanks so much for the update. I can take a closer look in a bit and let you know. Thanks again

oscb (approved):

alanjcharles:
@raminos we're ready to merge this as soon as the test you added passes! It looks pretty straightforward but let me know if I can help at all. Thanks!

alanjcharles:
accounted for in #436 thanks again!

PR #401, opened 2022-01-05
[Adjust] Call trackEvent in the track method

Kerumen:
The `trackEvent` call from Adjust was missing in the `track` method 🙈


jetaix (approved):

alanjcharles:
@Kerumen thanks so much for all of the help so far. If you would like to send me your email, I'd be happy to add you to our community slack so you have an easier way to reach us. If you don't want to add it here, email me at acharles@twilio.com. Will merge when tests pass

Issue #397, opened 2022-01-04
`device.id` is `null` on Android


Kerumen:
On Android, we have `null` for the `device.id` key.

```
"device": {
  "id": null,
  "manufacturer": "Google",
  "model": "Pixel 3",
  "name": "blueline",
  "type": "android"
},
```

I've tracked it down to the `collectDeviceId` key which is read in the Java implement of `getContextInfo` but not passed in the call on the JavaScript side:

https://github.com/segmentio/analytics-react-native/blob/829fc80bc8c4f59fa99dadd25083efa422d577f0/packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt#L101

https://github.com/segmentio/analytics-react-native/blob/829fc80bc8c4f59fa99dadd25083efa422d577f0/packages/core/src/context.ts#L34

2 questions:

- Why do we need to add a condition to get the `deviceId`?
- If the condition is necessary, why `getContextInfo` is not passing the key accordingly?

alanjcharles:
@Kerumen  i believe we made this optional (conditional) because Google has recently changed their consent policy so we can't collect it it automatically anymore. To answer your second question, looks like it might be another bug. Sorry you're running into so many today. I'll add it to the list for this week and keep you posted.

Kerumen:
@alanjcharles No worries! That's the purpose of a beta test. Happy to help, don't hesitate to ping me if you have something to try out.

PR #405, opened 2022-01-11
fix: add fix for deviceId on android builds

alanjcharles:
fix for #397

### DIFF
--- a/packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
+++ b/packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
@@ -76,7 +76,7 @@ class AnalyticsReactNativeModule(reactContext: ReactApplicationContext) : ReactC
     }
 
     @ReactMethod
-    fun getContextInfo(collectDeviceId: Boolean, promise: Promise) {
+    fun getContextInfo(config: ReadableMap, promise: Promise) {
       val appName: String = reactApplicationContext.applicationInfo.loadLabel(reactApplicationContext.packageManager).toString()
       val appVersion: String = pInfo.versionName
       val buildNumber = getBuildNumber()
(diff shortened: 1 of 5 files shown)


packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
```diff
@@ -76,7 +76,7 @@ class AnalyticsReactNativeModule(reactContext: ReactApplicationContext) : ReactC
     }
 
     @ReactMethod
-    fun getContextInfo(config: ReadableMap, promise: Promise) {
+    fun getContextInfo(collectDeviceId: Boolean, promise: Promise) {
```
oscb, line 79: We should keep the params as before so as to not differ from iOS. Also the config makes it more extendable.

packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
```diff
@@ -98,7 +98,7 @@ class AnalyticsReactNativeModule(reactContext: ReactApplicationContext) : ReactC
       contextInfo.putString("appVersion", appVersion)
       contextInfo.putString("buildNumber", buildNumber)
       contextInfo.putString("bundleId", bundleId)
-      contextInfo.putString("deviceId", getUniqueId(config.hasKey("collectDeviceId") && config.getBoolean("collectDeviceId")))
+      contextInfo.putString("deviceId", getUniqueId(collectDeviceId))
```
oscb, line 101: We should check the config as before to only request it when the config says so.

oscb (changes requested):

Kerumen (approved):

Kerumen (approved):

oscb (approved):

PR #413, opened 2022-01-25
fix: remove console.warning when sending events

oscb:


Kerumen:
Thanks for this, it was very annoying in development! 😄

PR #419, opened 2022-01-27
fix: support m1 macs in the example project

oscb:
Reverting to 0.63 in the example project seems to do the trick with the extra lines in the Podfile.
It is only the example project so it should be fine if we use an older RN version there, the main lib still supports newer RN versions.


alanjcharles:
I think we can close this one now right?

oscb:
> I think we can close this one now right?

Yea, I'll close this

PR #422, opened 2022-01-31
feature: replacing redux with sovran

oscb:
- fix: support m1 macs in the example project
- feat: replacing redux with sovran


packages/core/src/storage/sovranStorage.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+    this.userInfoStore = createStore(INITIAL_VALUES.userInfo, {
+      persist: { storeId: `${this.storeId}-userInfo` },
+    });
+  }
+
+  readonly isReady = {
+    get: () => true,
+    onChange: (_callback: (value: boolean) => void) => {
+      // Not doing anything cause we don't yet have a way to persist
```
alanjcharles, line 59: NIT we should remove this comment and maybe this function entirely?

oscb: Let's not for now, I think it might actually come in handy if we want to do some preprocessing or anything during initialization

osamakhn:
but why @alanjcharles? replace it with Recoil I or hooks I would say

alanjcharles:
> 

@osamakhn not sure I understand your question entirely. Why Sovran?

alanjcharles (approved):

PR #425, opened 2022-02-01
refactor: bump rn version in example app to resolve vulnerabilities

alanjcharles:


example/package.json
```diff
@@ -26,7 +26,7 @@
     "@react-navigation/stack": "^6.0.7",
     "@segment/sovran-react-native": "^0.2.0",
     "react": "17.0.2",
```
oscb, line 28: Does React need to be updated too?

oscb (approved):

example/package.json
alanjcharles: no, that's what the diff website is showing. Still trying to figure out what's going on with the Android test

PR #428, opened 2022-02-02
refactor: add ^ to react and react native

alanjcharles:


oscb (approved):

Kerumen:
@alanjcharles @oscb The same way we did for Sovran, I think this PR should be partially reverted to keep the `peerDependencies` but remove the explicit `dependencies`.

Wdyt?

PR #436, opened 2022-02-09
refactor: fix identify argument type and tests

alanjcharles:


packages/core/src/__tests__/methods/identify.test.ts
```diff
@@ -79,4 +79,23 @@ describe('methods #identify', () => {
...
   });
+
+  it('does not update userId when userId is undefined', () => {
+    const client = new SegmentClient(clientArgs);
+    jest.spyOn(client, 'process');
+
+    client.identify();
+
+    const expectedEvent = {
+      traits: initialUserInfo.traits,
```
oscb, line 90: `initialUserInfo` are the values already in the mock store, maybe use some other trait values so we can confirm it is changing them?

oscb (approved):

PR #450, opened 2022-02-22
refactor: fix malformed context

alanjcharles:
@oscb be sure to add a write-key to retrieve settings (i can dm you mine if needed) a test is currently failing too.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -33,7 +33,7 @@ import { IdfaPlugin } from '@segment/analytics-react-native-plugin-idfa';
 import { AmplitudeSessionPlugin } from '@segment/analytics-react-native-plugin-amplitude-session';
 
 const segmentClient = createClient({
-  writeKey: 'WRITE KEY',
+  writeKey: '',
   trackAppLifecycleEvents: true,
   collectDeviceId: true,
   debug: true,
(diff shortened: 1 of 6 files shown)


example/src/App.tsx
```diff
@@ -33,7 +33,7 @@ import { IdfaPlugin } from '@segment/analytics-react-native-plugin-idfa';
 import { AmplitudeSessionPlugin } from '@segment/analytics-react-native-plugin-amplitude-session';
 
 const segmentClient = createClient({
-  writeKey: 'WRITE KEY',
```
oscb, line 36: Let's keep this for the example

packages/core/src/timeline.ts
```diff
@@ -26,9 +26,22 @@ export class Timeline {
...
+          plugin.update({ integrations: newSettings }, UpdateType.initial);
+        }
+      });
     }
+
+    plugin.analytics?.settings.onChange((newSettings) => {
+      if (newSettings !== undefined) {
+        plugin.update({ integrations: newSettings }, UpdateType.refresh);
+      }
+    });
```
oscb, line 43: These 2 subscriptions should be combined into one. Right now it will call update twice (one with initial type and the other with refresh type) since it doesn't unsubscribe the initial one.

packages/core/src/timeline.ts
```diff
@@ -26,9 +26,22 @@ export class Timeline {
...
+    console.log('ADD PLUGIN', plugin);
```
oscb, line 44: Remove

oscb (approved):

PR #475, opened 2022-03-10
feat: add anonymousId logic to reset

alanjcharles:


packages/core/src/__tests__/analytics.test.ts
```diff
@@ -119,11 +119,24 @@ describe('SegmentClient', () => {
...
+        traits: undefined,
+      });
+    });
+
     it('resets user data, identity, traits', () => {
       client = new SegmentClient(clientArgs);
       const setUserInfo = jest.spyOn(store.userInfo, 'set');
 
-      client.reset();
+      client.reset(true);
```
oscb, line 139: For this test can we make it test the default behavior (`client.reset()`) so that we make sure we retain the same behavior for current users? Should only require to revert this line.

packages/core/src/analytics.ts
```diff
@@ -651,12 +651,18 @@ export class SegmentClient {
     this.appState = nextAppState;
   }
 
-  reset() {
+  reset(resetAnonymousId: Boolean = true) {
```
oscb, line 654: Uppercase `Boolean` in TS is an object type. It should be `boolean` to be the primitive.

oscb (changes requested):

oscb (approved):

PR #477, opened 2022-03-18
feat: added destination metadata and internal timeline processing for destinatio plugins

oscb:


packages/core/src/plugins/SegmentDestination.ts
```diff
@@ -1,53 +1,31 @@
...
-      plugins?.map((plugin) => (plugin as DestinationPlugin).key) ?? [];
-    const disabledCloudIntegrations: SegmentAPIIntegrations = {};
-    if (pluginSettings !== undefined) {
-      for (const key of deviceModePlugins) {
-        if (key in pluginSettings) {
-          disabledCloudIntegrations[key] = false;
-        }
-      }
+  execute(event: SegmentEvent): SegmentEvent | undefined {
+    const enrichedEvent = super.execute(event);
```
prayansh, line 24: will this trigger the full timeline?

packages/core/src/plugins/__tests__/SegmentDestination.test.ts
```diff
@@ -72,14 +76,57 @@ describe('SegmentDestination', () => {
...
+    expect(result).toEqual({
+      ...event,
+      _metadata: {
+        bundled: ['firebase'],
+        unbundled: [],
+        bundledIds: [],
+      },
+    });
+  });
+
```
prayansh, line 89: here is a more detailed test for the destination metadata plugin https://github.com/segmentio/analytics-kotlin/blob/main/core/src/test/kotlin/com/segment/analytics/kotlin/core/platform/plugins/DestinationMetadataPluginTests.kt#L70, if you wanted to expand this

packages/core/src/plugins/SegmentDestination.ts
oscb: Yes, it comes from `DestinationPlugin` superclass so that all destinations get it

PR #498, opened 2022-04-07
feat: add mixpanel plugin

alanjcharles:


packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  update(settings: SegmentAPISettings, _: UpdateType) {
+    const mixpanelSettings = settings.integrations[
+      this.key
+    ] as SegmentMixpanelSettings;
+
+    if (!mixpanelSettings || this.mixpanel !== undefined) {
+      return;
+    }
+    if (mixpanelSettings.token.length) {
```
oscb, line 36: strict check `> 0`

packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -0,0 +1,76 @@
...
+      return;
+    }
+
+    if (mixpanelSettings.enableEuropeanEndpoint) {
+      this.mixpanel?.setServerURL('api.eu.mixpanel.com');
+    }
+  }
+
+  identify(event: IdentifyEventType) {
+    identify(event, this.mixpanel!, this.settings!);
```
oscb, line 50: I don't think it is safe to force unwrap `this.mixpanel` and `this.settings` since they get assigned by the update only when the settings were received.

I think it should be handled here or in each method explicitely:
```
if (this.mixpanel !== undefined && this.settings !== undefined) {
  identify(event, this.mixpanel, this.settings)
}
return event
```

oscb: (Same for all the other methods)

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mockIntegrationSettings.ts
```diff
@@ -0,0 +1,18 @@
...
+      peopleProperties: [],
+      propIncrements: ['prop1', 'prop2', 'prop3'],
+      setAllTraitsByDefault: false,
+      superProperties: [],
+      trackAllPages: false,
+      trackCategorizedPages: false,
+      trackNamedPages: false,
+    },
+  },
+};
```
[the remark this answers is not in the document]
oscb: Also rename to `sampleIntegrationSettings` or something like that, since it is not a mock per se

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    store.reset();
+    jest.clearAllMocks();
+  });
+
+  it('calls update with settings', () => {
+    const plugin = new MixpanelPlugin();
+    plugin.analytics = new SegmentClient(clientArgs);
+    let settings = mockIntegrationSettings;
+    let updateType: UpdateType = 'initial' as UpdateType;
+    const mockMixpanelPlugin = jest.spyOn(plugin, 'update');
```
oscb, line 32: `updateSpy` seems to be a more appropriate name since it is a spy instead of a mock and also only spies on the update call of the plugin

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it('initializes Mixpanel', () => {
+    const plugin = new MixpanelPlugin();
+    plugin.analytics = new SegmentClient(clientArgs);
+    let settings = mockIntegrationSettings;
+    let updateType: UpdateType = 'initial' as UpdateType;
+
+    jest.spyOn(plugin, 'update');
+    plugin.update(settings, updateType);
+    expect(initMock).toHaveBeenCalled();
```
oscb, line 46: You can combine this with the test above. It would make test running faster given it's fewer tests!

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  beforeEach(() => {
+    store.reset();
+    jest.clearAllMocks();
+  });
+
+  it('calls update with settings', () => {
+    const plugin = new MixpanelPlugin();
+    plugin.analytics = new SegmentClient(clientArgs);
+    let settings = mockIntegrationSettings;
+    let updateType: UpdateType = 'initial' as UpdateType;
```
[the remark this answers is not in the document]
oscb: Also you might not actually need to store it in a variable, you can just pass the enum value directly in the call

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it('does not initialize Mixpanel when token is undefined', () => {
+    const plugin = new MixpanelPlugin();
+    plugin.analytics = new SegmentClient(clientArgs);
+    let settings = mockIntegrationSettings;
+    let updateType: UpdateType = 'initial' as UpdateType;
+
+    settings.integrations.Mixpanel.token = '';
+
+    jest.spyOn(plugin, 'update');
```
oscb, line 57: not needed

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    plugin.update(settings, updateType);
+    expect(initMock).toBeCalledTimes(0);
```
oscb, line 59: `expect(initMock).not.toHaveBeenCalled()` is the more jest-y way 😃

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  it('enables the European endpoint', () => {
+    const plugin = new MixpanelPlugin();
+    plugin.analytics = new SegmentClient(clientArgs);
+    let settings = mockIntegrationSettings;
+    let updateType: UpdateType = 'initial' as UpdateType;
+
+    settings.integrations.Mixpanel.token = '1234';
+    settings.integrations.Mixpanel.enableEuropeanEndpoint = true;
+
+    jest.spyOn(plugin, 'update');
```
oscb, line 71: not needed

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    plugin.update(settings, updateType);
+    expect(setServerMock).toHaveBeenCalled();
```
oscb, line 73: Also add a check to confirm it was called with the right value. Making the string a constant as suggested above will also be handy since you can just import the constant to confirm the value here without repeating strings

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mixpanel-react-native.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  init = initMock;
+  setServerURL = setServerMock;
+  setLoggingEnabled = jest.fn();
+  setUseIpAddressForGeolocation = jest.fn();
+  hasOptedOutTracking = jest.fn();
+  optInTracking = jest.fn();
+  optOutTracking = jest.fn();
+  identify = identifyMock;
+  alias = jest.fn();
```
oscb, line 24: `alias = aliasMock;` ?

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mixpanel-react-native.ts
```diff
@@ -0,0 +1,64 @@
...
+  track = jest.fn();
```
oscb, line 25: `track = trackMock;`

packages/plugins/plugin-mixpanel/src/methods/__tests__/alias.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      type: 'alias',
+      userId: '123',
+    } as AliasEventType;
+    const mixpanel = new Mixpanel('123');
+
+    alias(payload, mixpanel);
+
+    expect(mixpanel.getDistinctId).toBeCalled();
+    // need to fix this one
+    // expect(mixpanel.alias).toBeCalled();
```
oscb, line 21: I think it might have not worked due to the missing `aliasMock`

packages/plugins/plugin-mixpanel/src/methods/__tests__/group.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        newGroup: true,
+        coolGroup: false,
+        members: 4,
+      },
+      groupId: '23322',
+    } as GroupEventType;
+    const settings: SegmentMixpanelSettings =
+      mockIntegrationSettings.integrations.Mixpanel;
+    const mixpanel = new Mixpanel('1234');
+    const result = jest.spyOn(mixpanel, 'getGroup');
```
oscb, line 27: `const getGroupSpy =` to keep consistency

packages/plugins/plugin-mixpanel/src/methods/__tests__/group.test.ts
```diff
@@ -0,0 +1,55 @@
...
```
oscb, line 48: same as above

packages/plugins/plugin-mixpanel/src/methods/__tests__/identify.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  it('calls identify with userId', () => {
+    const payload = {
+      type: 'identify',
+      traits: {
+        firstName: 'John',
+        phone: '(555) 555-5555',
+        foo: 'bar',
+      },
+      userId: 'user',
+    } as IdentifyEventType;
```
oscb, line 22: nit: DRY by putting the default stuff in a const and using destructuring to replace the values you need for each test. Like:
```
const defaultPayload = {
        type: 'identify',
        traits: {
          firstName: 'John',
          phone: '(555) 555-5555',
          foo: 'bar',
        },
        userId: 'user',
      } as IdentifyEventType;
// ... then inside the test
const payload = { ...defaultPayload, userId: undefined }
```

packages/plugins/plugin-mixpanel/src/methods/__tests__/identify.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    const settings: SegmentMixpanelSettings =
+      mockIntegrationSettings.integrations.Mixpanel;
+    const mixpanel = new Mixpanel('1234');
+
+    identify(payload, mixpanel, settings);
+
+    expect(mixpanel.identify).toBeCalled();
```
oscb, line 29: For all these tests I think we should confirm that we are calling these methods in mixpanel with the right values, not only that the calls are done.
If the values are too complex to write here make use of snapshots: https://jestjs.io/docs/snapshot-testing

packages/plugins/plugin-mixpanel/src/methods/__tests__/identify.test.ts
```diff
@@ -0,0 +1,186 @@
...
+      },
+      userId: undefined,
+    } as IdentifyEventType;
+    const settings: SegmentMixpanelSettings =
+      mockIntegrationSettings.integrations.Mixpanel;
+    const mixpanel = new Mixpanel('1234');
+
+    identify(payload, mixpanel, settings);
+
+    expect(mixpanel.identify).toBeCalledTimes(0);
```
oscb, line 48: `.not.toHaveBeenCalled()`

packages/plugins/plugin-mixpanel/src/methods/__tests__/identify.test.ts
```diff
@@ -0,0 +1,186 @@
...
+        phone: '(555) 555-5555',
+        foo: 'bar',
+      },
+      userId: 'user',
+    } as IdentifyEventType;
+    const settings: SegmentMixpanelSettings =
+      mockIntegrationSettings.integrations.Mixpanel;
+    const mixpanel = new Mixpanel('1234');
+    settings.setAllTraitsByDefault = true;
+    const result = jest.spyOn(mixpanel, 'getPeople');
```
oscb, line 65: naming

oscb (changes requested):

packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -0,0 +1,76 @@
...
+    return event;
+  }
+
+  track(event: TrackEventType) {
+    const eventName = event.event;
+    const properties = event.properties as JsonMap;
+
+    mixpanelTrack(eventName, properties, this.settings!, this.mixpanel!);
```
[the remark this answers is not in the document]
alanjcharles: because it works for mixpanel's track and screen methods not just track mainly because of the people API.

packages/plugins/plugin-mixpanel/src/methods/__tests__/MixpanelPlugin.test.ts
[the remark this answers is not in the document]
alanjcharles: that's true, I do change values in it in other tests so I felt like it was just more consistent happy to discuss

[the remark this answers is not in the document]
alanjcharles: I can't seem to because everything loses its reference- happy to discuss to see what I'm doing wrong

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mockIntegrationSettings.ts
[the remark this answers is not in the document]
alanjcharles: changed the name and moved it into a different folder under `__mocks__` because I can't figure out how to get jest to ignore it in the `__tests__` folder.

packages/plugins/plugin-mixpanel/src/methods/__tests__/identify.test.ts
alanjcharles: I can do this once I figure out how to clean up the duplicated code with `beforeEach()`

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mockIntegrationSettings.ts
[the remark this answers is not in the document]
oscb: Ah, put it in `_tests__/__helpers__`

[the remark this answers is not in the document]
alanjcharles: yeah i did that and it still failed😅

packages/plugins/plugin-mixpanel/src/methods/group.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (groupId !== null && groupTraits?.length) {
+    for (let groupTrait of groupTraits) {
+      for (let eventTrait in event.traits) {
+        if (groupTrait.toLocaleLowerCase() === eventTrait.toLocaleLowerCase()) {
+          let group = event.traits[groupTrait] as string;
+          let traits = event.traits;
+
+          mixpanel.getGroup(group, groupId).setOnce('properties', traits);
+        }
+      }
```
oscb, line 24: Per our conversation, unless we know that the `groupTraits` and `eventTraits` casing might be mixed between them you could same some time and code processing this by just accessing directly by key.

```
for (const groupTrait of groupTraits) {
  const group = event.traits[groupTrait] as string;
  const traits = event.traits;

  mixpanel.getGroup(group, groupId).setOnce('properties', traits);
}
```

in another note, you can declare everything with `const` instead of `let` since nothing is changing.

packages/plugins/plugin-mixpanel/src/methods/identify.ts
```diff
@@ -0,0 +1,64 @@
...
+  lastName: '$last_name',
+  createdAt: '$created',
+  lastSeen: '$last_seen',
+  email: '$email',
+  name: '$name',
+  username: '$username',
+  phone: '$phone',
+} as any;
+
+const mapTransform = generateMapTransform(traitMap, {});
```
oscb, line 19: nice!

packages/plugins/plugin-mixpanel/src/methods/identify.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (settings.setAllTraitsByDefault === true) {
+    mixpanel.registerSuperProperties(mixpanelTraits);
+
+    if (settings.people === true) {
+      mixpanel.getPeople().set(mixpanelTraits);
+    }
+  }
+
+  if (settings.superProperties?.length) {
```
oscb, line 41: strict condition: `> 0`

packages/plugins/plugin-mixpanel/src/methods/identify.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    for (let superProperty of superProperties) {
+      superPropertyTraits[superProperty] = mixpanelTraits[superProperty];
+    }
+
+    let mappedSuperProperties = mapTransform(superPropertyTraits);
+    mixpanel.registerSuperProperties(mappedSuperProperties);
+  }
+
+  if (settings.people === true && settings.peopleProperties?.length) {
```
oscb, line 53: strict condition: ` > 0`

packages/plugins/plugin-mixpanel/src/methods/mixpanelTrack.ts
```diff
@@ -0,0 +1,56 @@
...
+        mixpanel.getPeople().increment(property, 1);
+
+        let lastEvent = `Last ${property}`;
+        let lastDate = Date();
+        mixpanel.getPeople().set(lastEvent, lastDate);
+      }
+    }
+  }
+
+  if (properties.revenue) {
```
oscb, line 51: strict condition: are we expecting this to be !undefined or > 0 ?

packages/plugins/plugin-mixpanel/src/methods/mixpanelTrack.ts
```diff
@@ -0,0 +1,56 @@
...
+          let incrementValue = properties[property];
+          if (typeof incrementValue === 'number') {
+            mixpanel.getPeople().increment(property, incrementValue);
+          }
+        }
+      }
+    }
+  }
+
+  if (settings.eventIncrements?.length) {
```
oscb, line 36: strict condition

packages/plugins/plugin-mixpanel/src/methods/mixpanelTrack.ts
```diff
@@ -0,0 +1,56 @@
...
+) => {
+  //track raw event
+  mixpanel.track(eventName, properties);
+
+  //everything else is for people setting
+  if (settings.people !== true) {
+    return;
+  }
+
+  if (settings.propIncrements?.length) {
```
oscb, line 21: strict condition

packages/plugins/plugin-mixpanel/src/methods/mixpanelTrack.ts
```diff
@@ -0,0 +1,56 @@
...
+      for (let property in properties) {
+        if (propString.toLowerCase() === property.toLowerCase()) {
+          let incrementValue = properties[property];
+          if (typeof incrementValue === 'number') {
+            mixpanel.getPeople().increment(property, incrementValue);
+          }
+        }
+      }
+    }
+  }
```
oscb, line 34: similar to the group comment, do we expect the casing to change in the keys between `propIncrements` and `properties`? If not we should just access items by the key and save cycles

packages/plugins/plugin-mixpanel/src/methods/mixpanelTrack.ts
```diff
@@ -0,0 +1,56 @@
...
+      if (eventString.toLowerCase() === eventName.toLowerCase()) {
+        let property = eventName;
+        mixpanel.getPeople().increment(property, 1);
+
+        let lastEvent = `Last ${property}`;
+        let lastDate = Date();
+        mixpanel.getPeople().set(lastEvent, lastDate);
+      }
+    }
+  }
```
oscb, line 49: nit: a bunch of `let` that should be `const`

packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+    return event;
+  }
+
+  alias(event: AliasEventType) {
+    if (this.mixpanel !== undefined) {
+      alias(event, this.mixpanel);
+    }
+    return event;
+  }
+}
```
oscb, line 87: doesn't this need a `flush` ? how are these events getting to mixpanel?
Looks like they have their own interval, but it would be good if calls to our `flush` also triggered the mixpanel flushing
https://developer.mixpanel.com/docs/react-native#flushing-events

oscb (changes requested):

packages/core/src/types.ts
```diff
@@ -239,6 +239,22 @@ export type SegmentAdjustSettings = {
...
+  trackNamedPages?: boolean;
+  trackCategorizedPages?: boolean;
+  people?: boolean;
+  setAllTraitsByDefault?: boolean;
+  superProperties?: string[];
+  peopleProperties?: string[];
+  groupIdentifierTraits?: string[];
+  eventIncrements?: string[];
+  propIncrements?: string[];
+};
```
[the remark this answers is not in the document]
alanjcharles: I put it in here because we have other Plugin types defined here like Adjust. Should we move them all?

packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
alanjcharles: ohh good shout

packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+import alias from './methods/alias';
+import mixpanelTrack from './methods/mixpanelTrack';
+
+export const EU_SERVER = 'api.eu.mixpanel.com';
+export class MixpanelPlugin extends DestinationPlugin {
+  type = PluginType.destination;
+  key = 'Mixpanel';
+  trackScreens = false;
+  private mixpanel: Mixpanel | undefined;
+  private settings: SegmentMixpanelSettings | undefined = undefined;
```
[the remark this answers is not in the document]
alanjcharles: same problem I'm having in Adjust

packages/plugins/plugin-mixpanel/src/methods/__mocks__/mixpanel-react-native.ts
```diff
@@ -0,0 +1,64 @@
+export const initMock = jest.fn();
+export const setServerMock = jest.fn();
+export const identifyMock = jest.fn();
+export const aliasMock = jest.fn();
+export const trackMock = jest.fn();
+export const trackWithGroupsMock = jest.fn();
+export const setGroupMock = jest.fn();
+
+export class Mixpanel {
```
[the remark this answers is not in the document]
alanjcharles: thank you :)

packages/plugins/plugin-mixpanel/src/methods/alias.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { AliasEventType } from '@segment/analytics-react-native';
+
+export default async (event: AliasEventType, mixpanel: Mixpanel) => {
+  let distinctId = '';
+  let newId = event.userId as string;
+
+  try {
+    distinctId = await mixpanel.getDistinctId();
+  } catch (e) {
+    console.log(e);
```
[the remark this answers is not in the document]
alanjcharles: definitely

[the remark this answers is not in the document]
alanjcharles: oh, it looks like we moved logger to the example plugin repo so I don't want to import it here. I'll make a note of that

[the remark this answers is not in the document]
oscb: I think that's a different Logger. We have this one in core still it should be accessible through the client: https://github.com/segmentio/analytics-react-native/blob/master/packages/core/src/analytics.ts#L50
you could do: `this.analytics.logger.warn(e)`

oscb (changes requested):
Just one more change about the logger and resurfacing the question about why the `mixpanelTrack` file isn't called just `track` like the rest.

oscb (approved):
Tests look great!

oscb (approved):

alanjcharles:
> 

thanks.! :)

PR #499, opened 2022-04-07
feat: add install attribution callback to adjust plugin

alanjcharles:
adds Adjust's `Attribution Callback` to track [`Install Attributed` ](https://github.com/adjust/react_native_sdk#attribution-callback)

### DIFF
--- a/packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
+++ b/packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
@@ -6,6 +6,7 @@ import {
   SegmentAdjustSettings,
   SegmentAPISettings,
   UpdateType,
+  EventType,
 } from '@segment/analytics-react-native';
 import { Adjust, AdjustConfig } from 'react-native-adjust';
 import identify from './methods/identify';
@@ -16,6 +17,7 @@ export class AdjustPlugin extends DestinationPlugin {
   type = PluginType.destination;
   key = 'Adjust';
 
+  private adjust: AdjustConfig | undefined = undefined;
   private settings: SegmentAdjustSettings | null = null;
 
   update(settings: SegmentAPISettings, _: UpdateType) {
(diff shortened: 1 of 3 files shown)


packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
```diff
@@ -16,9 +16,11 @@ export class AdjustPlugin extends DestinationPlugin {
   type = PluginType.destination;
   key = 'Adjust';
 
+  private adjust: AdjustConfig | undefined = undefined;
```
oscb, line 19: nit: `undefined` is the default, so no need to assign it

packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
```diff
@@ -16,9 +16,11 @@ export class AdjustPlugin extends DestinationPlugin {
...
   private settings: SegmentAdjustSettings | null = null;
 
   update(settings: SegmentAPISettings, _: UpdateType) {
+    const analytics = this.analytics ?? undefined;
```
oscb, line 23: I think the null-coalescing isn't necessary? if `this.analytics` is `undefined` it will remain `undefined`

packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
```diff
@@ -35,6 +37,24 @@ export class AdjustPlugin extends DestinationPlugin {
 
     const adjustConfig = new AdjustConfig(this.settings.appToken, environment);
 
+    if (this.adjust !== undefined) {
+      adjustConfig.setAttributionCallbackListener(function (attribution) {
```
oscb, line 41: oh, it's the first time I see this notation for an inline function, TIL!

oscb: Oh, I see why you created the `const analytics`, so that you captured the value of the analytics in the clousure.
You can preserve the `this` reference if you declare your function here with an arrow:
```
adjustConfig.setAttributionCallbackListener((attribution) => {
	...
	this.analytics?.track('Install Attributed', trackPayload); // Arrow functions closure on this automatically
});

oscb (changes requested):

packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
alanjcharles: ohhhh that is way simpler!!

packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
alanjcharles: Not sure I understand. Removing `undefined` gives me an error 

```Property 'adjust' has no initializer and is not definitely assigned in the constructor```

oscb (approved):
lgtm!

PR #502, opened 2022-04-13
feat: spin off the queue handling into a reusable plugin

oscb:


oscb:
Some important learnings:
- Jest `useFakeTimers` completely screws up the async calls which was the original problem with the tests: we need to use `jest.useFakeTimers('legacy')` or use another library for faking timers for now.
- Our API for analytics already stated that `flush` is awaitable, yet we're not awaiting for the plugins to respond, fixed that.
- RN doesn’t support `Promise.allSettled` which is pretty handy for not throwing errors when we have to await promises but don’t actually want to do anything in particular for errors (and don’t want to throw)

509dave16:
@oscb The QueueFlushingPlugin is not making use of the `storePersistor` client option :sweat_smile:. This would break Gopuff's current integration with Analytics React Native, since we rely on the that option so that we can use our MMKV persistor.

oscb:
Thanks for pointing that out. I'll add the option for it before we release this.

509dave16:
@oscb Thank you!

oscb:
Added the code to use the same persistor as in the config in #602

PR #513, opened 2022-04-25
feat: add reset to mixpanel plugin

alanjcharles:


packages/plugins/plugin-mixpanel/src/MixpanelPlugin.ts
```diff
@@ -91,4 +91,10 @@ export class MixpanelPlugin extends DestinationPlugin {
       this.mixpanel?.flush();
     }
   }
+
+  reset(): void {
+    if (this.isInitialized()) {
+      this.mixpanel?.reset();
```
oscb, line 97: since we already to the optional chaining do we need to check for `this.isInitialized()` ?

alanjcharles: good point! I guess we also don't need it for `flush()`

oscb (approved):

PR #517, opened 2022-04-27
Clarify Expo support

brentvatne:
Users can remain in the "managed workflow" (where they don't have `ios/` or `android/` directories in their project and only write JS) by using dev clients and EAS Build. The limitation that exists for this library is that it won't work in Expo Go.


alanjcharles:
thanks @brentvatne!

alanjcharles (approved):

PR #518, opened 2022-04-27
fix: fix callback conditional

alanjcharles:


packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
```diff
@@ -16,8 +16,8 @@ export class AdjustPlugin extends DestinationPlugin {
   type = PluginType.destination;
   key = 'Adjust';
 
-  private adjust: AdjustConfig | undefined;
   private settings: SegmentAdjustSettings | null = null;
+  private registeredCallback: Boolean = false;
```
oscb, line 20: Lets use `hasRegisteredCallback` to make it more aligned with the flags:

```suggestion
  private hasRegisteredCallback: Boolean = false;
```

packages/plugins/plugin-adjust/src/AdjustPlugin.tsx
```diff
@@ -66,6 +66,7 @@ export class AdjustPlugin extends DestinationPlugin {
         adjustConfig.setDelayStart(delayTime);
       }
     }
+    this.registeredCallback = true;
```
oscb, line 69: You need to put this inside the condition that checks if the callback has been registered (after registering the callback)

oscb (changes requested):

oscb (approved):

PR #528, opened 2022-05-06
feat: add attribution callbacks to appsflyer

alanjcharles:


packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
+      this.hasInitialized = true;
+    } else {
+      appsFlyer.initSdk({
+        devKey: this.settings.appsFlyerDevKey,
+        appId: this.settings.appleAppID,
+        ...defaultOpts,
+      });
+      this.hasInitialized = true;
+    }
   }
```
oscb, line 77: Remove

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
     track(event);
     return event;
   }
+
+  conversionCallback = () => {
+    appsFlyer.onInstallConversionData((res) => {
+      const isFirstLaunch = JSON.parse(res?.data?.is_first_launch);
+      const status = res?.data?.af_status;
```
oscb, line 92: Remove

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
 export class AppsflyerPlugin extends DestinationPlugin {
   type = PluginType.destination;
-  key = 'Appsflyer';
+  key = 'AppsFlyer';
 
-  constructor(opts: InitSDKOptions) {
-    super();
+  private settings: SegmentAppsflyerSettings | null = null;
+  private hasRegisteredInstallCallback: Boolean = false;
+  private hasRegisteredDeepLinkCallback: Boolean = false;
```
oscb, line 20: Should use lowercase `boolean` as the types with Uppercase are objects, not primitives

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
 
-    const defaultOpts = {
+  private hasInitialized: Boolean = false;
```
oscb, line 22: Also this one

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
 
-    appsFlyer.initSdk({
-      ...defaultOpts,
-      ...opts,
-    });
+    const appsflyerSettings = settings.integrations[
+      this.key
+    ] as SegmentAppsflyerSettings;
+
+    if (!appsflyerSettings) {
```
oscb, line 35: explicit check here: `appsflyerSettings !== undefined`

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
+      return;
+    }
+    const clientConfig = this.analytics?.getConfig();
+
+    this.settings = appsflyerSettings;
+
+    if (
+      this.settings.trackAttributionData === true &&
+      this.hasRegisteredInstallCallback === false
```
oscb, line 44: nit: No need for explicit checks in these ones as they are booleans.

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
     track(event);
     return event;
   }
+
+  conversionCallback = () => {
```
oscb, line 89: The name of this function should be like `registerConversionCallback` as this isn't the callback itself but a function that registers it.

It's usually a good idea to start all the functions with a verb

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
...
+        if (status === 'Non-organic') {
+          this.analytics?.track('Install Attributed', properties);
+        } else {
+          this.analytics?.track('Organic Install', { provider: 'AppsFlyer' });
+        }
+      }
+    });
+  };
+
+  deepLinkCallback = () => {
```
oscb, line 111: Same here, the name should be like `registerDeepLinkCallback`

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
     track(event);
     return event;
   }
+
+  conversionCallback = () => {
+    appsFlyer.onInstallConversionData((res) => {
+      const isFirstLaunch = JSON.parse(res?.data?.is_first_launch);
```
oscb, line 91: Why does it have to parse this variable? Is it a string instead of a boolean or something? can't we use it directly?

oscb: Asking because `JSON.parse` might be an expensive operation

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
...
+      const status = res?.data?.af_status;
+      const properties = {
+        provider: 'Appsflyer',
```
oscb, line 94: Should this match the key? maybe we should put all these into constants.

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
oscb: A good practice would also be to unwrap all the values you need from `res?.data` as in:
```
const { af_status,  media_source, is_first_launch } = res?.data;
```

You can also assign them default values when destructuring.

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -35,4 +85,41 @@ export class AppsflyerPlugin extends DestinationPlugin {
...
+        }
+      }
+    });
+  };
+
+  deepLinkCallback = () => {
+    appsFlyer.onAppOpenAttribution((res) => {
+      if (res?.status === 'success') {
+        const properties = {
+          provider: 'Appsflyer',
```
oscb, line 115: Same here, maybe we put this in a constant

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
+      this.hasInitialized === false &&
+      clientConfig?.trackDeepLinks === true
+    ) {
+      appsFlyer.initSdk({
+        devKey: this.settings.appsFlyerDevKey,
+        appId: this.settings.appleAppID,
+        onDeepLinkListener: true,
+        ...defaultOpts,
+      });
+      this.hasInitialized = true;
```
oscb, line 68: Were you able to confirm if `initSdk` can be called multiple times, cause if so it would be better since that would mean we can re-init it with new settings if that ever happens. If that's not possible lets add a comment with your findings.

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -3,27 +3,77 @@ import {
...
+    } else {
+      appsFlyer.initSdk({
+        devKey: this.settings.appsFlyerDevKey,
+        appId: this.settings.appleAppID,
+        ...defaultOpts,
+      });
+      this.hasInitialized = true;
+    }
```
oscb, line 76: This `else` needs a condition since this will execute every time, as `this.hasInitialized = true` gets here.

The correct way would be to check again `else if (!this.hasInitialized)`.

However there's only one param that changes when `clientConfig.trackDeepLinks` is `true` so this simplifies it:

```
if (!this.hasInitialized) {
  appsFlyer.initSdk({
    devKey: this.settings.appsFlyerDevKey,
    appId: this.settings.appleAppID,
    onDeepLinkListener: clientConfig?.trackDeepLinks === true,
    ...defaultOpts,
  });
  this.hasInitialized = true;
}
```

oscb (changes requested):

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
alanjcharles: it is possible to call `initSDK` multiple times per my testng

oscb (approved):

PR #547, opened 2022-05-18
fix: appsflyer attribution

Asgaroth:
Fixes AppsFlyer plugin not reporting install events due to the condition being always false:

Types in `appsflyer-react-native` package are wrongly reporting `is_first_launch: "true" | "false";`
But at runtime, these are actual boolean values.

https://github.com/AppsFlyerSDK/appsflyer-react-native-plugin/blob/master/index.d.ts#L12

Also added extra safety JSON.parse based on the examples in https://github.com/AppsFlyerSDK/appsflyer-react-native-plugin/blob/master/Docs/DeepLink.md#-1-deferred-deep-linking-get-conversion-data

Also, fixed types for `analytics.add(plugin)`

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -26,7 +26,6 @@ import {
   JsonMap,
   PluginType,
   SegmentAPIIntegrations,
-  SegmentAPISettings,
   SegmentEvent,
   UpdateType,
   UserInfoState,
@@ -360,13 +359,15 @@ export class SegmentClient {
   }: {
     plugin: P;
     settings?: P extends DestinationPlugin ? IntegrationSettings : never;
-
   }) {
     // plugins can either be added immediately or
     // can be cached and added later during the next state update
     // this is to avoid adding plugins before network requests made as part of setup have resolved
     if (settings !== undefined && plugin.type === PluginType.destination) {
-      this.store.settings.add((plugin as unknown as DestinationPlugin).key, settings);
+      this.store.settings.add(
+        (plugin as unknown as DestinationPlugin).key,
+        settings
+      );
     }
 
     if (!this.store.isReady.get()) {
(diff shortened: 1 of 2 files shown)


alanjcharles:
@Asgaroth thanks so much for this, looks great! I added one change request but happy to approve once that's sorted.

Asgaroth:
@alanjcharles did you maybe forget to submit? I don't see the changes requested.

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -85,7 +85,7 @@ export class AppsflyerPlugin extends DestinationPlugin {
         },
       };
 
-      if (is_first_launch === 'true') {
+      if (is_first_launch && JSON.parse(is_first_launch) === true) {
```
oscb, line 88: To prevent truthy values but safeguard agains the inconsistent types of AF SDK let's just use the JSON parse value.
```suggestion
      if (JSON.parse(is_first_launch) === true) {
```

packages/core/src/analytics.ts
```diff
@@ -351,20 +352,21 @@ export class SegmentClient {
 
   /**
    * Adds a new plugin to the currently loaded set.
-   * @param {{ plugin: Plugin, settings?: SegmentAPISettings }} Plugin to be added. Settings are optional if you want to force a configuration instead of the Segment Cloud received one
+   * @param {{ plugin: Plugin, settings?: IntegrationSettings }} Plugin to be added. Settings are optional if you want to force a configuration instead of the Segment Cloud received one
    */
-  add({
+  add<P extends Plugin>({
```
oscb, line 357: Thanks for this!

oscb (changes requested):

oscb (approved):

packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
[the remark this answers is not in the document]
alanjcharles: @Asgaroth

alanjcharles (approved):

PR #556, opened 2022-05-31
fix: appsflyer attribution error due to install conversion data failure

manueljtejada:
This PR prevents the `onInstallConversionData` handler from throwing an error when the SDK is not able to get the conversion data. In this case, `res?.data?.is_first_launch` would be undefined and `JSON.parse` would fail.

Sample payload in `onInstallConversionData` that would cause this error:

```
{
  status: 'failure',
  type: 'onInstallConversionFailure',
  data: 'Launch status code: 404'
}
```


alanjcharles (approved):

alanjcharles:
thanks @manueljtejada!

PR #567, opened 2022-06-10
fix: refactor getDeviceID to use DRM, remove AndroidID

oscb:


a613:
Is there any chance this change results in a higher number of device IDs being reported, or duplicates, or anything weird like that? After upgrading from 1.5.0 to 2.2.2, we have seen an increase in the number of device IDs from segment, and worse, even though we pass the segment anonymous ID to AppsFlyer, the IDs coming from both systems do not align anymore.

Looking for any ideas you might have, because something serious happened at the time of that upgrade. Thanks.

oscb:
@a613 there has been a few fixes in the deviceId logic since 2.2.2, in particular thinking that you might be needing this fix: https://github.com/segmentio/analytics-react-native/releases/tag/%40segment%2Fanalytics-react-native-v2.6.0

Are you on the latest? If you do please open a new bug with more details for us to check it out!

Issue #520, opened 2022-04-29
proxy missing in the setup config


jayasme:
## Summary

Before version 2.0 it has a proxy property in the type `Configuration` and received by the `setup` function, so I can set the proxy info to the Segment, but after I upgraded the version to 2.x, the `proxy` is no longer supported, however, this is an important function I need, is there a way to add the proxy to Segment?
I can't find why it has been removed from `README.md` or `CHANGELOG.md` even `CHANGELOG.md` has been removed as well, nor in the migration guide, nor in the documentation.

## Proposed Solution

Bring `proxy` back! THANK YOU!

509dave16:
## Proposed Solution
The following changes would need to be done at minimum in order for Analytics React Native to support the `proxy` client option again:
- Update `Client Options` section table in `README.md` to include a row for `proxy` which will be a `string`
- Update `Config` typescript interface for client options argument passed to `createClient` to include `proxy` as a `string`
- Update `sendEvents.ts` to reference `proxy` from `config` parameter

509dave16:
@jayasme The `proxy` client option should be available in the next release. Will probably happen later this week. Hope that helps!

oscb:
This is now released! https://github.com/segmentio/analytics-react-native/releases/tag/%40segment%2Fanalytics-react-native-v2.3.1

PR #570, opened 2022-06-13
feat: Create Client Proxy Option

509dave16:
# Summary
The gist of this PR is that it allows for a `proxy` client option to be specified in the `createClient` function call. Previously in Analytics React Native 1.0, a `proxy` client option was available in the `setup` function call.

# Use Case
Gopuff's Customer Data Platform team wants to be able to empower other internal engineerings teams with the ability to inspect events and assert them for integrations tests in realtime. To do this we were hoping to be able to specify a `proxy` that the events would be posted to first before we post them to Segment. `analytics-node` and `analytics-next` already have the ability to specify proxies, so we were hoping we could make it possible for `analytics-react-native`.

# Documentation
Updated README.md's section on **Client Options**. Hopefully that was sufficient. Please let me know if additional documentation is required.

# Unit Tests
Updated `api.test.ts` to include a test case that asserts that the `proxy` client option overrides the Segment batch url.

# References
- Feature request: https://github.com/segmentio/analytics-react-native/issues/520


oscb (approved):
This is great!
We had this in our sights but  hadn't found the time to take care of it, thanks for the help!

509dave16:
@oscb No worries bud! Really appreciate you taking the time to review, approve, and merge!!

PR #574, opened 2022-06-16
Feat: Add AdvertisingId Plugin

alanjcharles:
Adds an AdvertisingId Plugin to collect the `advertisingId` from Google Play Services on Android devices


oscb (changes requested):

example/android/app/build.gradle
```diff
@@ -1,5 +1,5 @@
 apply plugin: "com.android.application"
-apply plugin: 'com.google.gms.google-services'
+// apply plugin: 'com.google.gms.google-services'
```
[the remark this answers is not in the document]
alanjcharles: I added the dependencies back but we can't use this plugin without a `google-services.json` 
image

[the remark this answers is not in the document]
oscb: I'm curious why is it doing this task `app:processDebugGoogleServices` since there's no Gradle plugin anymore. Do we need the google gradle plugin in the `android-kotlin` lib?

packages/plugins/plugin-advertising-id/android/src/main/java/com/reactnativeanalyticsreactnativepluginadvertisingid/AnalyticsReactNativePluginAdvertisingIdModule.kt
```diff
@@ -0,0 +1,41 @@
...
+    ?.reactNativeHost
+    ?.reactInstanceManager
+    ?.currentReactContext
+
+    if (reactContext == null) {
+      promise.resolve(null)
+      return
+    }
+
+     val advertisingInfo = AdvertisingIdClient.getAdvertisingIdInfo(reactContext)
```
oscb, line 31: Per the docs:

> Cases where Google Play Services is unavailable are indicated by a [GooglePlayServicesNotAvailableException](https://developers.google.com/android/reference/com/google/android/gms/common/GooglePlayServicesNotAvailableException) being thrown by getAdvertisingIdInfo().

We should handle exceptions here, log it and return null in those cases

alanjcharles:
@oscb I can't get it to work with the CI for the reasons we discussed on Slack. I'm pretty much out of ideas, so happy to discuss. Otherwise, I will go ahead and update the readme with the google items/deps  a user will have to add/implement and go from there. Thanks dude!

packages/core/android/google-services.json
```diff
@@ -10,7 +10,7 @@
       "client_info": {
         "mobilesdk_app_id": "1:169668264691:android:3c64b2cb29f7ff3516ed86",
         "android_client_info": {
-          "package_name": "com.example.segmentanalyticsreactnative"
+          "package_name": "com.segmentanalyticsreactnative"
```
oscb, line 13: why is this file in `core`? it contains data particular to your test instance right? we should remove it.

oscb (changes requested):

.github/workflows/ci.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         with:
           api-level: ${{ matrix.api-level }}
           profile: ${{matrix.profile}}
-          avd-name: Pixel_API_21_AOSP
+          avd-name: Pixel_API_31_AOSP
```
oscb, line 131:
```suggestion
          avd-name: Pixel_API_21_AOSP
```

oscb (approved):

PR #579, opened 2022-06-24
fix: map correct facebook events

Lorex7:
This PR:

- map the missing Checkout Started [spec’d events](https://segment.com/docs/connections/spec/) to the matching Facebook Event following the [docs](https://segment.com/docs/connections/destinations/catalog/facebook-app-events/#facebook-events).
- correctly trigger a facebook purchase event just in case the event is a `fb_mobile_purchase`. Previously the event was triggered also for events having the props `_valueToSum`, e.g. `Product Added`


oscb (approved):
This looks good, thanks for the PR!

PR #581, opened 2022-06-24
fix: persist user traits across events

oscb:
- chore: enable semantic-release
- fix: persist users traits across events


alanjcharles (approved):
lgtm!

Issue #572, opened 2022-06-15
Wrong path to package.json for commonjs


jeremybarbet:
- `analytics-react-native` version: 2.2.2
- Integrations versions (if used):
- React Native version: 0.67.4
- iOS or Android or both? both

**Steps to reproduce**

- Follow this step https://github.com/segmentio/analytics-react-native#setting-up-the-client
- Running the app results in a module not found error `../package.json in context.js`

**Expected behavior**

- To find the `package.json` file

**Actual behavior**

After bundled, `context.js` is in `lib/commonjs/context.js` rather than in `src/context.ts`. The path to the `package.json` is then wrong

That's the patch I'm using until the path can be determined based on the root directory:

```patch
diff --git a/node_modules/@segment/analytics-react-native/lib/commonjs/context.js b/node_modules/@segment/analytics-react-native/lib/commonjs/context.js
index 19e2207..b8fbea4 100644
--- a/node_modules/@segment/analytics-react-native/lib/commonjs/context.js
+++ b/node_modules/@segment/analytics-react-native/lib/commonjs/context.js
@@ -7,7 +7,7 @@ exports.getContext = void 0;
 
 var _reactNative = require("react-native");
 
-var _package = _interopRequireDefault(require("../package.json"));
+var _package = _interopRequireDefault(require("../../package.json"));
 
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```

jeffcongdon:
experiencing this same behavior outside of commonjs, running test suite in jest.  Using version `2.2.2`, but confirmed error goes back to `2.0.0-pilot`

oscb:
This should be fixed in `2.4.0`

PR #594, opened 2022-07-07
Fix/package path

NoRoboto:
Based on https://github.com/segmentio/analytics-react-native/issues/572

This solution implements the creation of a `info.ts` file so as not to use `package.json` at runtime since it is outside of `src` folder, making it unreachable for invocations.

Why:

- Don't use the whole package just the necessary variables.
- Copying the entire file is not such a good solution.
- `process.env.npm_package_version` is not a solution because this is not Node runtime.

Downsides:

- Each version change in the `package.json` generates a new change to commit in the new file.

It's not the most elegant solution, but it works! :p, I would like to know your thoughts on this.

Thank you!

### DIFF
--- a/packages/core/constants-generator.js
+++ b/packages/core/constants-generator.js
@@ -1,0 +1,17 @@
+const fs = require('fs');
+const packageJson = require('./package.json');
+
+const body = `
+export const libraryInfo = {
+  name: '${packageJson.name}',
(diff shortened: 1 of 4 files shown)


oscb:
I think this will work for now, thanks for taking this one up.

oscb (approved):

PR #596, opened 2022-07-11
fix: map to correct firebase event name/attributes

Lorex7:
This PR:
Correctly map the following event names (the name previously used are now [deprecated](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#public-static-final-string-ecommerce_purchase)):
- `Order Complete` maps to `purchase` instead of `ecommerce_purchase`
- `Promotion Viewed` maps to `view_promotion` instead of `present_offer`
-  `Order Refunded` maps to `refund` instead of `purchase_refund`

We also added a map for the events props inspired by the facebook event plugin, as some events were not fired if the attributes provided didn't have the correct names.


oscb (approved):

oscb:
Thanks for the fix here! I will release shortly

PR #600, opened 2022-07-13
feat: support max kB in upload batching

oscb:


prayansh (approved):

packages/core/src/util.ts
```diff
@@ -1,18 +1,43 @@
...
 
-  let index = 0;
-  let resIndex = 0;
-  const result = new Array(Math.ceil(array.length / size));
+  let currentChunk = 0;
+  let rollingKBSize = 0;
+  const result: any[] = array.reduce(
+    (chunks: any[][], item: any, index: number) => {
+      if (maxKB !== undefined) {
+        rollingKBSize += sizeOf(item);
```
prayansh, line 19: wont this be more accurate to do in plain bytes rather than kb? we wont have to do the `/ 1024` everytime we add bytes that way

PR #604, opened 2022-07-15
fix: timeline safe processing events for destinations

oscb:
-  Timeline won't shortcircuit destinations if one fails or returns undefined (each DestinationPlugin should be independent)
- Fixed Braze plugin key ('Appboy')

### DIFF
--- a/packages/core/src/plugin.ts
+++ b/packages/core/src/plugin.ts
@@ -155,7 +155,7 @@ export class DestinationPlugin extends EventPlugin {
 
   execute(event: SegmentEvent): SegmentEvent | undefined {
     if (!this.isEnabled(event)) {
-      return undefined;
+      return;
     }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/plugin.ts
```diff
@@ -155,7 +155,7 @@ export class DestinationPlugin extends EventPlugin {
 
   execute(event: SegmentEvent): SegmentEvent | undefined {
     if (!this.isEnabled(event)) {
-      return undefined;
+      return;
```
prayansh, line 158: whats the difference here?

prayansh (approved):
LGTM

packages/core/src/plugin.ts
oscb: Nothing. Just for consistency.

PR #616, opened 2022-07-26
refactor: add userInfo.onChange to listen for new user traits

alanjcharles:


packages/plugins/plugin-braze/src/BrazePlugin.tsx
```diff
@@ -13,22 +13,25 @@ export class BrazePlugin extends DestinationPlugin {
   key = 'Appboy';
 
   identify(event: IdentifyEventType) {
-    const currentUserInfo = this.analytics?.userInfo.get();
+    let currentUserInfo = null;
```
oscb, line 16: use `undefined` instead of `null`

packages/plugins/plugin-braze/src/BrazePlugin.tsx
```diff
@@ -13,22 +13,25 @@ export class BrazePlugin extends DestinationPlugin {
...
+
+    if (currentUserInfo === null) {
       let integrations = event.integrations;
 
       if (integrations !== undefined) {
         integrations[this.key] = false;
       }
     } else {
       identify(event);
+      currentUserInfo = null;
```
oscb, line 34: `undefined`

packages/plugins/plugin-braze/src/BrazePlugin.tsx
```diff
@@ -13,22 +13,25 @@ export class BrazePlugin extends DestinationPlugin {
...
-    if (
-      currentUserInfo?.userId === event.userId &&
-      currentUserInfo?.anonymousId === event.anonymousId &&
-      currentUserInfo?.traits === event.traits
-    ) {
+    this.analytics?.userInfo.onChange((newUserInfo) => {
+      if (newUserInfo !== undefined) {
+        currentUserInfo = newUserInfo;
+      }
+    });
```
oscb, line 24: This is subscribing to any changes on every `identify` call, which isn't the intention. This isn't actually working as you are thinking, cause `currentUserInfo` is never set in the outer scope, so it never executes the `identify` call to Braze down below. (It is always going to be `null`)

I think we might have to change this whole logic. We cannot use the `onChange` subscription cause using it would mean to either subscribe outside of the timeline execution (hence not here), and lose the timeline integration. It should be some rolling variable, you store only the last value seen for `userInfo` on an `identify` event. Pretty much all the logic willbe the same, you check if the upcoming event has any changes from what you last seen and trigger Braze accordingly

oscb (changes requested):

PR #618, opened 2022-07-27
Clarify FaceBook Plugin Docs

niallzato:
Add more clarity for how customers initialize the Plugin and what methods they need to use.


alanjcharles (approved):
lgtm!

PR #622, opened 2022-07-28
refactor: update braze identify debounce logic

alanjcharles:


oscb (changes requested):

packages/plugins/plugin-braze-middleware/README.md
```diff
@@ -0,0 +1,67 @@
+# @segment/analytics-react-native-plugin-braze-middleware
+
+`BeforePlugin`  to debounce `identify` events for [Braze](https://www.braze.com).
```
oscb, line 3: Let's add that this is for people who are using `Braze` through cloud mode, and also have a link for the Braze device mode plugin here. (and do the reverse in the Device Mode plugin)

packages/plugins/plugin-braze-middleware/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "lib",
+    "android",
+    "ios",
+    "cpp",
+    "segment-analytics-react-native.podspec",
+    "package.json",
+    "!src/**/*.e2e.mock.js",
+    "!**/__tests__",
+    "!**/__fixtures__",
+    "!**/__mocks__"
```
oscb, line 34: add also here the `!android/.gradle` and the other one that you just added in `core`

packages/plugins/plugin-braze-middleware/README.md
oscb: Also add that you cannot have both in the same instance.

packages/plugins/plugin-braze/src/BrazePlugin.tsx
```diff
@@ -11,24 +12,24 @@ import flush from './methods/flush';
 export class BrazePlugin extends DestinationPlugin {
   type = PluginType.destination;
   key = 'Appboy';
+  private lastSeenTraits: UserInfoState | undefined = undefined;
```
oscb, line 15: no need to initialize to `undefined`. that's the default

oscb (changes requested):

oscb (approved):

PR #624, opened 2022-08-01
Add information about `Application Installed` in 2.x migration

nchase:
`Application Installed` will re-fire after an upgrade from 1.x to 2.x, and this seems to be expected.

This information, along with some strategies for filtering out the extra events, is present in [your documentation at segment.com](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/migration/), but it's missing here. Its absence made it extremely difficult and scary to figure out why our `Application Installed` events were firing so frequently.

I added what was in segment.com's migration documentation. Alternatively, this file could mostly go away – you could link to the migration documentation on segment.com from here instead 🤷


alanjcharles:
thanks @nchase! Apologies for the confusion

alanjcharles (approved):

adaddeo:
We were also confused by this, thanks for adding. Is there any more information on creating an Enrichment Plugin or filtering in Segment workspace? I don't understand how to do either of those.

Andrerm124:
Has there been any more information provided into how to actually filter out these unwanted events?
It seems strange how it's a serious enough side effect to warrant a warning in the first paragraph of the Readme, but doesn't actually provide any useful information into how to actually resolve it.

PR #630, opened 2022-08-08
Set minSdkVersion to 19 per documentation for plugin-advertising-id

jedashford:
The current minSdkVersion set at 31 is breaking builds for anyone supporting older devices.
https://developers.google.com/android/guides/setup

To test your app when using Google Play services, you must use one of the following:

A compatible Android device that runs Android 4.4 (API level 19) or higher and has the Google Play Store app installed.
The Android emulator with an AVD that runs the Google APIs platform based on Android 4.4 (API level 19) or higher.


alanjcharles:
thanks for contributing @jedashford - will take a look tomorrow

oscb (approved):

PR #637, opened 2022-08-09
fix: fix device.id logic to prevent random ids

alanjcharles:


packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
```diff
@@ -139,16 +148,17 @@ class AnalyticsReactNativeModule : ReactContextBaseJavaModule, ActivityEventList
 
     val contextInfo: WritableMap = Arguments.createMap()
 
-    // generate random identifier that does not persist across installations
-    // use it as the fallback in case DRM API failed to generate one.
-    val fallbackDeviceId = UUID.randomUUID().toString()
-    val deviceId = getUniqueId(config.hasKey("collectDeviceId") && config.getBoolean("collectDeviceId"))?: fallbackDeviceId
+    val deviceId = getUniqueId(config.hasKey("collectDeviceId") && config.getBoolean("collectDeviceId"))?: getFallbackDeviceId(config.hasKey("collectDeviceId") && config.getBoolean("collectDeviceId"))
```
oscb, line 151: This seems too convoluted. Maybe just do a check if `config.hasKey("collectDeviceId")` and trigger the logic of getting the UniqueID or the fallback value. No need to repeat checking the config for every function.

alanjcharles: @oscb fair point. I just added another commit that works when i toggle the `collectDeviceId` config boolean. Let me know what you think

packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
```diff
@@ -76,7 +76,9 @@ class AnalyticsReactNativeModule : ReactContextBaseJavaModule, ActivityEventList
         md.update(wideVineId)
         return md.digest().toHexString()
       } catch (e: Exception) {
-        return null
+        // generate random identifier that does not persist across installations
+        // use it as the fallback in case DRM API failed to generate one.
+          return UUID.randomUUID().toString()
```
oscb, line 81: This wasn't exactly what I meant but this also works. Adjust the method signature then since this always returns a String not nullable

alanjcharles: @oscb it's still possible for this method to return null- 

``` if (collectDeviceId) {
      if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2)
        return null
```

I also have the logic here to account for the null possibilities: 

``` if (deviceId !== null) {
      contextInfo.putString("deviceId", deviceId)
    }
```

Thoughts?

oscb: It's weird that we handle differently wether we can't generate a code because of a failure with MediaDRM or because you are on an old version of Android.

I think it would be better to have the same fallback for both. 

This function should always return null if it can't create it, handle in the caller the scenario where this returns null and `collectDeviceId` is set

oscb (approved):

PR #660, opened 2022-09-01
fix: apply userInfo on processing to prevent concurrency issues

oscb:


packages/core/src/__tests__/__helpers__/mockSegmentStore.ts
```diff
@@ -16,15 +16,15 @@ import type {
...
   isReady: boolean;
   context?: DeepPartial<Context>;
   settings: SegmentAPIIntegrations;
   userInfo: UserInfoState;
   deepLinkData: DeepLinkData;
 };
 
-const INITIAL_VALUES: Data = {
+const INITIAL_VALUES: StoreData = {
   isReady: true,
```
509dave16, line 28: @oscb Shouldn't `isReady` initially be `false`?

oscb: I need to remove these, we don't use them anymore (we have more granular readiness checks now). Good catch.

Issue #605, opened 2022-07-15
feat: logLevel or logMessageCallback Client Option


509dave16:
## Summary

Our Data Dog client that’s integrated with the Gopuff app does not collect "warn" logs, which is is the log level Analytics React Native is using when there’s a failure to post a batch of events to Segment: https://github.com/segmentio/analytics-react-native/blob/55d798821163d5a41902a6bc099b1bfcbd853a17/packages/core/src/plugins/SegmentDestination.ts#L39

We cannot even begin to understand the errors that are causing the batches of events to fail posting to Segment going off of these `console.error`s alone:

![Screen Shot 2022-07-15 at 12 28 31 PM](https://user-images.githubusercontent.com/2925882/179299538-81b363b3-a702-4927-bbfa-1b4e410ca876.png)

## Proposed Solution

In order to for our Data Dog client to be able to capture these logs we need either a:
- `logLevel` client option to be able to specify "error" as the  log level.
- Or a `logMessageCallback` client option that we use in conjunction with the Data Dog client to explicitly capture these log messages

konoufo:
Hi @509dave16 . I think #670 now solves this for you. What do you think ?

509dave16:
@konoufo Yes! This does!! Appreciate your work here :)

PR #670, opened 2022-09-09
feat: adds option to provide a custom logger object to consume library logs

konoufo:
In response to issue #605 , this PR allows users to configure a custom logger via `createClient(config: Config)` as in:
```
const client = createClient({
   writeKey: 'SEGMENT_KEY',
   logger: customLogger
});
```
The custom logger just has to implement a simple interface `DeactivableLoggerType`.
This allows full flexibility to escalate logs, process them etc. And would effectively solve #605.


oscb (approved):
lgtm!

PR #673, opened 2022-09-13
Idfa fix

alanjcharles:
fix: add ability to enable/disable idfa plugin

### DIFF
--- a/packages/plugins/plugin-idfa/README.md
+++ b/packages/plugins/plugin-idfa/README.md
@@ -66,7 +66,7 @@ tracking permission and returns a promise you can use to
 make additional tracking decisions based on the response 
 */
 idfaPlugin.requestTrackingPermission().then((enabled: boolean) => {
-  console.log('IDFA -->', enabled);
+  console.log('Tracking Enabled -->', enabled);
 });
 ```
 ## Support
--- a/packages/plugins/plugin-idfa/README.md
+++ b/packages/plugins/plugin-idfa/README.md
@@ -51,14 +51,21 @@ const segmentClient = createClient({
 
 ...
 
-//The IDFA Plugin supports an optional `enabled` boolean 
-//setting to false disables plugin on initialization
+ /** The IDFA Plugin supports an optional `shouldAskPermission` boolean
+ which defaults to true. Setting to false prevents the plugin from 
+ requesting permission from the user. If you set the parameter to `false` on
+ initialization you **must** call `requestTrackingPermission()` 
+ to retrieve the `idfa`  
+ */
 const idfaPlugin = new IdfaPlugin(false);
 segmentClient.add({ plugin: idfaPlugin });
 
-//enable returns a promise you can use to 
-//make additional tracking decisions
-idfaPlugin.enable().then((enabled: boolean) => {
+
+/** `requestTrackingPermission()` will prompt the user for 
+tracking permission and returns a promise you can use to 
+make additional tracking decisions based on the response 
+*/
+idfaPlugin.requestTrackingPermission().then((enabled: boolean) => {
   console.log('IDFA -->', enabled);
 });
 ```
--- a/packages/plugins/plugin-idfa/README.md
+++ b/packages/plugins/plugin-idfa/README.md
@@ -36,6 +36,32 @@ const segmentClient = createClient({
 segmentClient.add({ plugin: new IdfaPlugin() });
 ```
 
+## Customize IDFA Plugin Initialization
+
+To delay the `IDFA Plugin` initialization (ie. to avoid race condition with push notification prompt) implement the following: 
+
+```ts
+import { createClient } from '@segment/analytics-react-native';
+
+import { IdfaPlugin } from '@segment/analytics-react-native-plugin-idfa';
+
+const segmentClient = createClient({
+  writeKey: 'SEGMENT_KEY'
+});
+
+...
+
+//The IDFA Plugin supports an optional `enabled` boolean 
+//setting to false disables plugin on initialization
+const idfaPlugin = new IdfaPlugin(false);
+segmentClient.add({ plugin: idfaPlugin });
+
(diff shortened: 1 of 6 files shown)


packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
```diff
@@ -3,36 +3,64 @@ import {
...
-import { IdfaEvents } from './IdfaEvents';
 import type { IdfaData } from './types';
 import { AnalyticsReactNativePluginIdfa } from './AnalyticsReactNativePluginIdfa';
 
 const { getTrackingAuthorizationStatus } = AnalyticsReactNativePluginIdfa;
 
 export class IdfaPlugin extends Plugin {
   type = PluginType.enrichment;
+  private isEnabled?: boolean = false;
+  private isDisabled?: boolean;
```
oscb, line 14: This is very confusing. I don't think also `isEnabled` needs to be an optional if it is always initialized to `false`

I think it would make more sense first by making them all positive boolean (`isEnabled`) instead of negative ones (`isDisabled`). Then renaming them:

- isDisabled -> shouldAskPermissions
- isEnabled -> isPermissionEnabled

That would make it easier to understand what each one does.

packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
```diff
@@ -3,36 +3,64 @@ import {
...
-    // subscribe to IDFAQuery event
-    // emitted when we prompt a user for permission
-    IdfaEvents.addListener('IDFAQuery', (res) => {
+    // since configure can be called multiple times potentially
+    // this accounts for both enabling and disabling plugin
+    if (this.isEnabled === false && this.isDisabled !== true) {
       this.getTrackingStatus();
-      this.analytics?.track('IDFAQuery', res);
-    });
+    }
```
oscb, line 34: We shouldn't ask for permissions on `configure`. This is only for Destination plugins that get new settings, but these settings are only set programatically, furthermore these are private so the user cannot change them afterwards.

packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
```diff
@@ -3,36 +3,64 @@ import {
...
+    } catch (error) {
+      this.analytics?.logger.warn(error);
+      return false;
+    }
+  }
+
+  //not sure we actually need this
+  disable() {
+    this.isDisabled = true;
   }
```
oscb, line 52: I don't think we need it. It is a one time operation, either you initialize this plugin and make it ask for permissions automatically or you don't and handle it yourself.

packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
```diff
@@ -3,36 +3,64 @@ import {
...
+    // since configure can be called multiple times potentially
+    // this accounts for both enabling and disabling plugin
+    if (this.isEnabled === false && this.isDisabled !== true) {
       this.getTrackingStatus();
-      this.analytics?.track('IDFAQuery', res);
-    });
+    }
+  }
+
+  async enable(): Promise<boolean> {
```
oscb, line 37: nit: maybe rename to `requestTrackingPermission` ?

oscb: Add some JSDoc comments too since this is a public method

packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
```diff
@@ -3,36 +3,64 @@ import {
...
 import { AnalyticsReactNativePluginIdfa } from './AnalyticsReactNativePluginIdfa';
 
 const { getTrackingAuthorizationStatus } = AnalyticsReactNativePluginIdfa;
 
 export class IdfaPlugin extends Plugin {
   type = PluginType.enrichment;
+  private isEnabled?: boolean = false;
+  private isDisabled?: boolean;
+
+  constructor(enabled?: boolean) {
```
oscb, line 16: Add some JSDoc comments as this constructor receives an additional parameter that should be documented.
I might also suggest renaming from `enabled` to `shouldAskPermissions` or something similar as I explained above.

oscb: Also the constructor in my opinion should default to `true`, which is kinda like the behavior it already has but more explicit.

packages/plugins/plugin-idfa/README.md
```diff
@@ -36,6 +36,32 @@ const segmentClient = createClient({
...
+import { IdfaPlugin } from '@segment/analytics-react-native-plugin-idfa';
+
+const segmentClient = createClient({
+  writeKey: 'SEGMENT_KEY'
+});
+
+...
+
+//The IDFA Plugin supports an optional `enabled` boolean 
+//setting to false disables plugin on initialization
```
oscb, line 55: A more true explanation would be that setting to `false` makes it not request automatically permissions, however it is super important to make it explicit that if they want IDFA to show up they absolutely need to call `.enable` (or the renamed function, see comment below). Else the IDFA won't retrieve the values for IDFA.

We should also add that by default it will request IDFA permissions right away.

packages/plugins/plugin-idfa/README.md
```diff
@@ -36,6 +36,32 @@ const segmentClient = createClient({
...
+const idfaPlugin = new IdfaPlugin(false);
+segmentClient.add({ plugin: idfaPlugin });
+
+//enable returns a promise you can use to 
+//make additional tracking decisions
+idfaPlugin.enable().then((enabled: boolean) => {
```
oscb, line 61: We should also add that `.enable` will prompt the user for permissions (if not already granted) or update the values if already granted/denied.

oscb (changes requested):

oscb (approved):

PR #676, opened 2022-09-14
fix: avoids calling Firebase::setUserId with undefined value

konoufo:
This fixes a bug where calling `identify` on the client without specifying a `userId` could result in the Firebase Plugin throwing out type errors.


alanjcharles:
@konoufo looks great! do you still plan to add `reset()` functionality as well?

oscb (approved):
Great! Thanks for adding tests

PR #680, opened 2022-09-22
feat: short-circuit and throw warnings when missing native modules

konoufo:


packages/plugins/plugin-advertising-id/src/AdvertisingIdPlugin.ts
```diff
@@ -4,25 +4,29 @@ import {
...
+          } else {
+            this.setContext(id);
+          }
         }
-      }
-    );
+      );
+    } else {
+      warnMissingNativeModule();
+    }
```
oscb, line 29: nit: how about we have a function in `utils` for `getAnalyticsNativeModule` or something like that that either returns the module or `undefined` and triggers the `warnMissingNativeModule`

That way you can simplify the rest of the calls as:

```ts
getAnalyticsNativeModule()?.getAdvertisingId().then(
```

and 

```ts
getAnalyticsNativeModule()?.getContextInfo() ?? DefaultContext;
```

packages/plugins/plugin-advertising-id/src/AdvertisingIdPlugin.ts
```diff
@@ -4,25 +4,24 @@ import {
   PluginType,
   SegmentClient,
 } from '@segment/analytics-react-native';
-import { NativeModules } from 'react-native';
+import { getNativeModule } from '@segment/analytics-react-native/src/util';
```
oscb, line 7: Can we export this from the root so that the import is only `@segment/analytics-react-native`?

The reason this is important is because the releases contain CommonJS, Module and TS so that people can use it regardless of their import types. If we use `/src` it forces them to compile TS

konoufo: Oh I see. Got it. Will fix in a sec.

konoufo:
> Everything looks great! Just an important thing about the imports.

fixed

oscb (approved):

Issue #677, opened 2022-09-16
android crashes when setAttributionData is called


kevinbror:
Versions: 
react-native: ^0.67.2
@segment/analytics-react-native: ^2.2.1
@segment/analytics-react-native-plugin-appsflyer: ^0.4.0
@segment/analytics-react-native-plugin-braze: ^0.3.0
@segment/analytics-react-native-plugin-idfa: ^0.4.1

We recently updated from @segment/analytics-react-native-plugin-braze: 0.2.1 to 0.3.0 and @segment/analytics-react-native-plugin-appsflyer: from 0.3.2 to 0.4.0 and we are seeing periodic Android crashes `NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter adGroup`

We haven't been able to reproduce yet but checking the code it looks like maybe [this](https://github.com/segmentio/analytics-react-native/blob/%40segment/analytics-react-native-plugin-braze-v0.3.0/packages/plugins/plugin-braze/src/methods/track.ts#L12) is getting called with a bad value. 

Looking at the diff between versions is seems maybe [this](https://github.com/segmentio/analytics-react-native/compare/%40segment/analytics-react-native-plugin-appsflyer-v0.3.2...%40segment/analytics-react-native-plugin-appsflyer-v0.4.0#diff-b9768e366a9bffeef71de9e8581f4a4b667dcaf25f2ca747d35b9391a2f3cf5bR84) is the bug. track is called here without ad_group.

PR #681, opened 2022-09-22
fix: add null/undefined check to attribution data

alanjcharles:
- adds safety check for attribution values in `Install Attributed` events in Braze Plugin

### DIFF
--- a/packages/plugins/plugin-braze/src/methods/track.ts
+++ b/packages/plugins/plugin-braze/src/methods/track.ts
@@ -5,6 +5,11 @@ export default (payload: TrackEventType) => {
   if (payload.event === 'Install Attributed') {
     if (payload.properties?.campaign) {
       const attributionData: any = payload.properties.campaign;
+      for (const [attribution, data] of Object.entries(attributionData)) {
+        if (data === undefined || data === null) {
+          attributionData[attribution] = '';
+        }
+      }
       const network = attributionData.source;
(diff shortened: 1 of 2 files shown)


packages/plugins/plugin-braze/src/methods/track.ts
```diff
@@ -5,6 +5,11 @@ export default (payload: TrackEventType) => {
   if (payload.event === 'Install Attributed') {
     if (payload.properties?.campaign) {
       const attributionData: any = payload.properties.campaign;
+      for (const [attribution, data] of Object.entries(attributionData)) {
+        if (data === undefined || data === null) {
+          attributionData[attribution] = '';
+        }
+      }
```
oscb, line 12: I don't think this code will handle `undefined` correctly because `Object.entries` will not return the key for those elements (they are essentially non-existent)

Maybe try a different approach using null-coalescing on the properties down below and having an constant object with initial values for each of them so that we can tweak them individually if needed.

Also add a test for both `null` and `undefined` scenarios! (specially the kind of `undefined` where you don't explicitely set it as `undefined`)

oscb (changes requested):
Left a single comment about an edge case this won't cover. It would be good if you can add tests for this as well!

oscb (approved):
lgtm!

alanjcharles:
#677

PR #694, opened 2022-10-12
Update min sdk version

TrekSoft:
In 1.0.2 this was updated to only require a min version of 19


oscb:
Thanks for this change!

oscb (approved):

PR #700, opened 2022-11-01
fix: remove special anonymousID behaviour on dev mode

### LINKED ISSUE #696: Incompatible anonymous ids
Latest segment version generates anonymous id which is not compatible with anonymous id of MixPanel.

image

oscb:
Fixes #696


alanjcharles (approved):
lgtm!

Issue #705, opened 2022-11-10
Error when run in Debug with Chrome on iOS Simulator


Manouli:
I am using the RN debugger to debug my app. The package seems to work fine when not debugging. When I debug with chrome I am getting the below error. 

![image](https://user-images.githubusercontent.com/6332957/201074993-02ab741f-5f8b-4f7f-9cc2-ef90c463d891.png)

This is happening when I trigger the createClient function.

- React Native version:  0.67.4
- iOS or Android or both? Yes
"@segment/analytics-react-native": "^2.9.1",
"@segment/sovran-react-native": "^0.4.5",

**Steps to reproduce**
Install the package and try to run it with the Debug with Chrome on.  

**Expected behavior**
It works with the debugger on.

**Actual behavior**
Its Breaks

rizshivalli:
+1

AdeenUlHaq3:
I also have this issue.

MorenoMdz:
Same issue

MorenoMdz:
It works if you downgrade Segment to 2.9.0

itajenglish:
+1 On this issue on React Native version: `0.70.5`. Downgrading to segment to `2.9.0` doesn't work for me.

itajenglish:
> It works if you downgrade Segment to 2.9.0

Just FYI for anyone downgrading. You must be on [`segment/sovran-react-native`](https://github.com/segmentio/sovran-react-native) version: `0.2.8`, not the latest version `0.4.5` for it to work.

PR #715, opened 2022-11-21
fix: add return value to uuid in dev mode

alanjcharles:
- addresses #705


oscb:
I think this works short term but we will start getting bugs about the AnonymousId being different during Debug again he he. 

A better long term solution is to make the `getUUIDSync` native method not being synchronous, since it has problems with old versions of RN and make it always async.

PR #716, opened 2022-11-22
Add `defaultSettings`'s type to documentation

luisfmsouza:
Include a better description of the expected value on the `defaultSettings` prop by adding a link to its type.

**Why?**

We had some trouble figuring it out, we had to open the source code and debug it to discover what value is expected for `defaultSettings`. Adding the type to the doc is a cheap way to prevent others going into the same trouble.


alanjcharles:
@luisfmsouza thank you!

alanjcharles (approved):

Issue #706, opened 2022-11-11
Error 'analytics-react-native' can't access a custom native module' after 2.9.1 upgrade


MorenoMdz:
- `analytics-react-native` version: 2.9.1
- React Native version: Expo SDK46 RN 0.69
- iOS or Android or both? iOS

**Steps to reproduce**

Load the app with the said version of Segment

**Expected behavior**

Segment to load correctly

**Actual behavior**
System shows warnings about Segment not being loaded:

```
iOS Bundling complete 3061ms
 WARN  Constants.platform.ios.model has been deprecated in favor of expo-device's Device.modelName property. This API will be removed in SDK 45.
 WARN  AsyncStorage has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-async-storage/async-storage' instead of 'react-native'. See https://github.com/react-native-async-storage/async-storage
 WARN  Constants.platform.ios.model has been deprecated in favor of expo-device's Device.modelName property. This API will be removed in SDK 45.
 LOG  ...Initializing Segment in development mode...
 WARN  The package 'analytics-react-native' can't access a custom native module. Make sure: 

- You have run 'pod install'
- You rebuilt the app after installing the package
- You are not using Expo managed workflow
```

myrs:
@MorenoMdz Hi!
I had a similar issue when using [`@segment/analytics-react-native-plugin-advertising-id`](https://github.com/segmentio/analytics-react-native/tree/master/packages/plugins/plugin-idfa) on `iOS`.

My solution was to conditionally add `AdvertisingIdPlugin` only on Android:
```
if (Platform.OS === 'android') {
  segmentClient.add({ plugin: new AdvertisingIdPlugin() });
}
```

My best guess is [this change](https://github.com/segmentio/analytics-react-native/compare/%40segment/analytics-react-native-v2.9.0...%40segment/analytics-react-native-v2.9.1#diff-b0b7eb5a1e896409f4d0e0cacdc260a7869dfcf9c000180cda61e72e4ada1b97R3) in `2.9.1` resulted in new warning, as previously it was ignored on `_DEV_`

MorenoMdz:
> @MorenoMdz Hi! I had a similar issue when using [`@segment/analytics-react-native-plugin-advertising-id`](https://github.com/segmentio/analytics-react-native/tree/master/packages/plugins/plugin-idfa) on `iOS`.
> 
[...]

PR #718, opened 2022-11-22
fix: replace native modules with react-native-uuid

alanjcharles:


packages/core/src/__tests__/uuid.test.ts
```diff
@@ -1,24 +1,11 @@
...
-
-    const result = getUUID();
+  it('should get a UUID from react-native-uuid', () => {
+    const uuidSpy = jest.spyOn(uuid, 'v4');
 
-    expect(result).toBe(id);
-    (global as any).__DEV__ = oldDevValue;
+    getUUID();
+    expect(uuidSpy).toBeCalled();
   });
```
oscb, line 10: I think we can remove this test. Not sure what it was trying to accomplish besides making sure we were calling a NativeModule method.

oscb (approved):
Looks good! Props for the fast turnaround on this.
I left a comment about a test we can remove imo, but otherwise this looks perfect

packages/core/src/__tests__/uuid.test.ts
alanjcharles: awesome thank you! I kinda thought so too haha. I'll remove it right now then.

alanjcharles:
#706

SchneiderOr:
Hey why this isn't reflected in master 2.9.1? i still getting this error on dev

PR #746, opened 2023-01-23
fix: refactor allSettled to remove dependencies

KonstantinosKoulaxizis:
Remove allSettled package method and use a local one with the same functionality.

allSettled method was throwing `Unhandled promise rejection [RangeError: Maximum call stack size exceeded.]`

### DIFF
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -94,3 +94,36 @@ export const getPluginsWithReset = (timeline: Timeline) => {
 
   return eventPlugins;
 };
+
+type PromiseResult<T> =
+  | {
+      status: 'fulfilled';
+      value: T;
+    }
+  | {
+      status: 'rejected';
+      reason: unknown;
+    };
+
+const settlePromise = async <T>(
+  promise: Promise<T> | T
+): Promise<PromiseResult<T>> => {
+  try {
+    const result = await promise;
+    return {
+      status: 'fulfilled',
+      value: result,
+    };
+  } catch (error) {
+    return {
+      status: 'rejected',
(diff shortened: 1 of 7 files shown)


oscb:
@KonstantinosKoulaxizis Thanks for the contribution! I don't quite get a repro on this error but it seems like a fairly small change so if this fixes it let's do it. I refactored the implementation a bit and added tests.

silesky (approved):
Coincidence... I recently wrote the same polyfill =)
```ts
type AllSettledResult = {
  status: 'fulfilled' | 'rejected'
  reason?: Error
  value?: T
}

// promise.allSettled "polyfill"
export const allSettled = (
  promises: Promise[]
): Promise[]> => {
  return Promise.all(
    promises.map((promise) => {
      // Per spec, Promise.allSettled([Promise.reject('oops'), 2] ->
      // [{status: "rejected", reason: "oops"}, { "status": "fulfilled", "value": 2 }}
      return Promise.resolve(promise)
        .then(
          (value) =>
            ({
              value,
              status: 'fulfilled',
            } as AllSettledResult)
        )
        .catch(
          (reason) =>
            ({
              reason,
              status: 'rejected',
            } as AllSettledResult)
        )
    })
  )
}


```

packages/core/src/util.ts
```diff
@@ -94,3 +94,36 @@ export const getPluginsWithReset = (timeline: Timeline) => {
 
   return eventPlugins;
 };
+
+type PromiseResult<T> =
```
silesky, line 98: This is an excellent typescript definition

PR #756, opened 2023-02-02
feat: add unified DL support to appsflyer

alanjcharles:
- adds support for AppsFlyer's new `onDeepLink()` method to support unified deeplinking
- bumps AppsFlyer SDK to `6.9.4`

### DIFF
--- a/packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
+++ b/packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
@@ -18,7 +18,6 @@ export class AppsflyerPlugin extends DestinationPlugin {
   private settings: SegmentAppsflyerSettings | null = null;
   private hasRegisteredInstallCallback: boolean = false;
   private hasRegisteredDeepLinkCallback: boolean = false;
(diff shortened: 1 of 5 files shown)


packages/plugins/plugin-appsflyer/src/AppsflyerPlugin.tsx
```diff
@@ -47,11 +48,15 @@ export class AppsflyerPlugin extends DestinationPlugin {
...
-      !this.hasRegisteredDeepLinkCallback
+      (clientConfig?.trackDeepLinks === true &&
+        !this.hasRegisteredDeepLinkCallback) ||
+      !this.hasRegisteredUnifiedLinkCallback
     ) {
       this.registerDeepLinkCallback();
       this.hasRegisteredDeepLinkCallback = true;
+
+      this.registerUnifiedDeepLinkCallback();
+      this.hasRegisteredUnifiedLinkCallback = true;
```
oscb, line 59: We don't need another flag for this one do we? Since they both get added at the same time

oscb (changes requested):
Just a single comment about a flag that might be unnecessary. Everything else looks good.

oscb (approved):

PR #759, opened 2023-02-09
feat: add clevertap plugin

alanjcharles:


packages/plugins/plugin-clevertap/src/ClevertapPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+import { mapTraits, transformMap } from './parameterMapping';
+import CleverTap from 'clevertap-react-native';
+const mappedTraits = generateMapTransform(mapTraits, transformMap);
+
+export class ClevertapPlugin extends DestinationPlugin {
+  type = PluginType.destination;
+  key = 'CleverTap';
+
+  identify(event: IdentifyEventType) {
+    const traits = event.traits as Record<string, any>;
```
oscb, line 19: Rather than `any` prefer `unknown`

oscb (changes requested):

packages/plugins/plugin-clevertap/src/ClevertapPlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    if (event.traits?.birthday !== undefined) {
+      const birthday = new Date(event.traits.birthday);
+      if (
+        birthday !== undefined &&
+        birthday !== null &&
+        !isNaN(birthday.getTime())
+      ) {
+        const data = new Date(event.traits.birthday);
+        console.log('DATEEE', data);
```
oscb, line 31: Remove

packages/plugins/plugin-clevertap/src/ClevertapPlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+export class ClevertapPlugin extends DestinationPlugin {
+  type = PluginType.destination;
+  key = 'CleverTap';
+
+  identify(event: IdentifyEventType) {
+    const traits = event.traits as Record<string, unknown>;
+    const safeTraits = mappedTraits(traits);
+    const userId = event.userId ?? event.anonymousId;
+
+    if (event.traits?.birthday !== undefined) {
```
oscb, line 23: Maybe try switching to using `safeTraits.DOB` everywhere since it's kinda confusing that it is reading the event but deleting the property in the `DOB`

oscb (changes requested):

oscb (approved):

PR #763, opened 2023-02-21
fix: add logic to convert traits to strings

ashton-huxtable:


packages/plugins/plugin-firebase/src/methods/identify.ts
```diff
@@ -6,6 +6,13 @@ export default async (event: IdentifyEventType) => {
     await firebaseAnalytics().setUserId(event.userId!);
   }
   if (event.traits) {
```
oscb, line 8: I think this might need to be done deeply: some of the traits are embedded objects so the "stringify" should apply to the properties of those objects too and so on 

Might be worth splitting this into its own `stringifyObject` function that gets called recursively if it finds a property in an object is another object.

oscb (changes requested):

packages/plugins/plugin-firebase/src/methods/identify.ts
```diff
@@ -6,6 +6,27 @@ export default async (event: IdentifyEventType) => {
...
+    let eventTraits = event.traits;
+    let safeTraits: Record<string, string>;
+
+    const checkType = (value: any) => {
+      return typeof value === 'object' && !Array.isArray(value);
+    };
+    safeTraits = Object.keys(eventTraits).reduce(
+      (acc: Record<string, string>, trait) => {
+        if (checkType(eventTraits[trait])) {
+          console.log(
```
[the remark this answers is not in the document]
oscb: Also this code path should return the `acc` right away so that it doesn't try to add it below

packages/plugins/plugin-firebase/src/methods/identify.ts
```diff
@@ -6,6 +6,27 @@ export default async (event: IdentifyEventType) => {
...
+            'We detected an object or array data type. Firebase does not accept nested traits.'
+          );
+        }
+        if (trait !== undefined) {
+          acc[trait] = eventTraits[trait]!.toString();
```
oscb, line 23: for safety it might be good not to force unwrap `eventTraits[trait]`. Technically it wouldn't ever happen since you are iterating over keys but it's a good practice for maintainability:
```
if (trait !== undefined && eventTraits[trait] !== undefined) {
  ...
```

packages/plugins/plugin-firebase/src/methods/identify.ts
```diff
@@ -6,6 +6,27 @@ export default async (event: IdentifyEventType) => {
     await firebaseAnalytics().setUserId(event.userId!);
   }
   if (event.traits) {
-    await firebaseAnalytics().setUserProperties(event.traits as any);
+    let eventTraits = event.traits;
+    let safeTraits: Record<string, string>;
+
+    const checkType = (value: any) => {
+      return typeof value === 'object' && !Array.isArray(value);
```
oscb, line 13: Shouldn't it be `typeof value === 'object' || Array.isArray(value);` ? so that it returns if the variable is array or object?

oscb (changes requested):

packages/plugins/plugin-firebase/src/methods/identify.ts
alanjcharles: there was a type error here until so i recommended force unwrapping since we can assume it will always be there (and said it's not good practice in most cases)

oscb (approved):

packages/plugins/plugin-firebase/src/FirebasePlugin.tsx
```diff
@@ -5,17 +5,45 @@ import {
...
+      let safeTraits: Record<string, string>;
+
+      const checkType = (value: unknown) => {
+        return typeof value === 'object' && !Array.isArray(value);
+      };
+      safeTraits = Object.keys(eventTraits).reduce(
+        (acc: Record<string, string>, trait) => {
+          if (checkType(eventTraits[trait])) {
+            this.analytics?.logger.warn(
+              'We detected an object or array data type. Firebase does not accept nested traits.'
```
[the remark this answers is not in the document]
alanjcharles: oh yeah, nice catch!

packages/plugins/plugin-firebase/src/FirebasePlugin.tsx
```diff
@@ -5,17 +5,45 @@ import {
...
 
-  identify(event: IdentifyEventType) {
-    identify(event);
+  async identify(event: IdentifyEventType) {
+    if (event.userId !== undefined) {
+      await firebaseAnalytics().setUserId(event.userId!);
+    }
+    if (event.traits) {
+      let eventTraits = event.traits;
+      let safeTraits: Record<string, string>;
```
silesky, line 23: why make this a separate declaration? =)

PR #775, opened 2023-03-07
feat: telemetry

oscb:
- feat: telemetry object on client, basic instrumentation
- feat: telemetry object on client, basic instrumentation

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -499,6 +499,7 @@ export class SegmentClient {
 
     await this.process(event);
     this.logger.info('SCREEN event saved', event);
+    this.telemetry.record('Screen');
   }
 
   async track(eventName: string, options?: JsonMap) {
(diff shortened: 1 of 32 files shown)


packages/core/src/telemetry/metrics.ts
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * Limit number of tags accepted by the server
+ */
+const TAGS_LIMIT = 10;
+
+/**
+ * We prioritize these tags over user defined ones if the number is over the limit, ordered by importance
+ */
+const PRIORITED_TAGS = [
```
alanjcharles, line 13: should this be `PRIORITIZED_TAGS`?

packages/core/src/analytics.ts
```diff
@@ -226,6 +234,8 @@ export class SegmentClient {
 
     // set up tracking for lifecycle events
     this.setupLifecycleEvents();
+
+    this.telemetry.record('Client Constructor', { test: 'test-OB' });
```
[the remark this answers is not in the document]
oscb: Yes I will remove

packages/core/src/telemetry/metrics.ts
oscb: Yup, must've been editing this way too late

PR #777, opened 2023-03-08
feat: add device token plugin

alanjcharles:


packages/plugins/plugin-device-token/src/methods/___tests__/DeviceTokenPlugin.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    let analytics = new SegmentClient(clientArgs);
+
+    jest.mock('react-native/Libraries/Utilities/Platform', () => ({
+      OS: 'ios',
+      select: () => null,
+    }));
+
+    await plugin.configure(analytics);
+
+    expect(configureSpy).toHaveBeenCalledWith(analytics);
```
oscb, line 46: this always succeeds

packages/plugins/plugin-device-token/src/methods/___tests__/DeviceTokenPlugin.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it('requests authorization when configure is called', async () => {
+    let configureSpy = jest.spyOn(plugin, 'configure');
+    let analytics = new SegmentClient(clientArgs);
+
+    jest.mock('react-native/Libraries/Utilities/Platform', () => ({
+      OS: 'ios',
+      select: () => null,
+    }));
```
oscb, line 42: Mocks should be at the top of the file because jest will actually execute them first (regardless of where you call them) (aka "hoist" them).

Also, I think we already have a global mock for `react-native`. Not sure if it has already a `Platform` object mocked but you can extend it

packages/plugins/plugin-device-token/src/methods/___tests__/DeviceTokenPlugin.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it('retrieves the device token when updatePermissions is called', async () => {
+    Platform.OS = 'ios';
+
+    await plugin.updatePermissionStatus();
+
+    expect(mockGetAPNSToken).toHaveBeenCalled();
+  });
+});
```
oscb, line 82: I think one important test missing here is to actually check that the side effects of getting a value (setting token inside context, a track call) are actually triggered.

You can either add that as a separate test or add it inside the tests you already have.

oscb (changes requested):

packages/plugins/plugin-device-token/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "directory": "packages/core"
+  },
+  "author": "Segment <hello@segment.com> (https://segment.com/)",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/segmentio/analytics-react-native/issues"
+  },
+  "homepage": "https://github.com/segmentio/analytics-react-native/tree/master/packages/plugins/plugin-firebase#readme",
+  "peerDependencies": {
+    "@react-native-firebase/app": "17.3.2",
```
oscb, line 48: it shouldn't be such an specific version, I think `17.x` is fine, we only need to stick it to a major version.

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,58 @@
...
+      let APNSToken = await messaging().getAPNSToken();
+      if (APNSToken !== null) {
+        await this.analytics?.context.set({ device: { token: APNSToken } });
+        this.analytics?.track('Device Token Retrieved');
+      }
+    } else if (Platform.OS === 'android') {
+      let deviceToken = await messaging().getToken();
+      if (deviceToken !== undefined && deviceToken.length) {
+        await this.analytics?.context.set({ device: { token: deviceToken } });
+        this.analytics?.track('Device Token Retrieved');
```
oscb, line 38: can we simplify this code? it's pretty much the same for both platforms, just the method that we call seems different, we can probably wrap that

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const isAuthorized = await this.authStatus;
+
+    if (isAuthorized) {
+      this.retrieveDeviceToken();
+    }
+  }
+}
+
+async function checkUserPermission() {
+  return await messaging().hasPermission();
```
oscb, line 57: Put a try/catch here, this might fail if the Native Modules have an issue I bet

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,58 @@
...
+      } else {
+        this.analytics?.logger.warn(
+          'Device token only available on iOS and Android platforms'
+        );
+      }
+    }
+  }
+
+  async updatePermissionStatus() {
+    const isAuthorized = await this.authStatus;
```
oscb, line 48: This won't work on demand, the `authStatus` promise is already resolved since it is the "result" of the function you called at constructiontime, You have to call `checkUserPermission` again.
Promises are just results, not functions.

packages/plugins/plugin-device-token/src/methods/___tests__/DeviceTokenPlugin.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    await plugin.updatePermissionStatus();
+
+    expect(mockGetAPNSToken).toHaveBeenCalled();
+  });
+
+  it('sets the device token in context for iOS', async () => {
+    let analytics = new SegmentClient(clientArgs);
+    await plugin.configure(analytics);
+
+    expect(store.context.get()).toEqual({
```
oscb, line 78: prefer to use the async safe version because this one might not actually resolve before the value is in there.
Also prefer to just check the actual token value, since this might fail if the context has more values

oscb:
```
const token = await store.context.get(true);
expect(token).toEqual('device-token');
```

packages/plugins/plugin-device-token/src/methods/___tests__/DeviceTokenPlugin.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  it('sets the device token in context for Android', async () => {
+    Platform.OS = 'android';
+    let analytics = new SegmentClient(clientArgs);
+    await plugin.configure(analytics);
+
+    expect(store.context.get()).toEqual({
+      device: {
+        token: 'device-token',
+      },
+    });
```
oscb, line 94: Same as above

oscb (changes requested):

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+      }
+    } else if (Platform.OS === 'android') {
+      let deviceToken = await messaging().getToken();
+      if (deviceToken !== undefined && deviceToken.length) {
+        this.setDeviceToken(deviceToken);
+      } else {
+        this.analytics?.logger.warn(
+          'Device token only available on iOS and Android platforms'
+        );
+      }
```
oscb, line 37: This `else` should be in the outer `if`

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+      this.analytics?.logger.warn('Not authorized to retrieve device token');
+    }
+  }
+
+  async retrieveDeviceToken() {
+    if (Platform.OS === 'ios') {
+      let APNSToken = await messaging().getAPNSToken();
+      if (APNSToken !== null) {
+        this.setDeviceToken(APNSToken);
+      }
```
oscb, line 28: You can go this way but there's even an extra way you can simplify this. I would actually have done it "backwards" by making the "reusable" function not setting the device token but instead one for getting the device token value in a standard way:

```
private async function getDeviceToken(): string | undefined {
  if (Platform.OS == 'ios') {
    return await messaging().getAPNSToken() ?? undefined; // This way we standard nulls and undefined into undefined only
  }
  if (Platform.OS == 'android') {
    let deviceToken = await messaging().getToken() ?? undefined;
    if (deviceToken !== undefined && deviceToken.length > 0) {
      return deviceToken;
    } else {
      return undefined;
    }
  }
  logger.warn("not available");
  return undefined;
}
```

This has the advantage that now your platform specific code is just one pure function without side effects. In theory you could just unit test this function for platform code and mock this one so that you don't have to mock platform on the rest.

Second advantage is that the output values now are normalized regardless of platform so now your `retrieveDeviceToken()` only needs to call this one and function normally:

```
let token = await getDeviceToken();
if (token !== undefined) {
  await context.set({ device: { token: token }});
  // etc
}
```

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+    this.analytics = analytics;
+
+    if (isAuthorized) {
+      this.retrieveDeviceToken();
+    } else {
+      this.analytics?.logger.warn('Not authorized to retrieve device token');
+    }
+  }
+
+  async retrieveDeviceToken() {
```
oscb, line 23: mark private

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+  PluginType,
+} from '@segment/analytics-react-native';
+import { Platform } from 'react-native';
+import messaging from '@react-native-firebase/messaging';
+
+export class DeviceTokenPlugin extends PlatformPlugin {
+  type = PluginType.enrichment;
+
+  async configure(analytics: SegmentClient) {
+    const isAuthorized = await this.checkUserPermission();
```
oscb, line 13: You can still do this with the promise variable you had before, that was still a good idea as the value is retrieved on construction time, then just awaited on `configure`.
By my previous comment I just meant that promises are a value. So you they won't change, but that still makes it useful here.

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+      this.retrieveDeviceToken();
+    }
+  }
+
+  async checkUserPermission() {
+    try {
+      return await messaging().hasPermission();
+    } catch (e) {
+      this.analytics?.logger.warn(e);
+      return;
```
oscb, line 59: nit: You can remove this return

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  async updatePermissionStatus() {
+    const isAuthorized = await this.checkUserPermission();
+
+    if (isAuthorized) {
+      this.retrieveDeviceToken();
+    }
+  }
+
+  async checkUserPermission() {
```
oscb, line 54: mark as private

oscb (changes requested):

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
alanjcharles: ohhh that makes a lot of sense. I updated it a bit differently but I think it works!

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
alanjcharles: it gives a blue linting error that not all code paths return a value so I added a return type

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
alanjcharles: oh okay! That makes complete sense thank you

packages/plugins/plugin-device-token/README.md
```diff
@@ -0,0 +1,99 @@
+# @segment/analytics-react-native-plugin-device-token
+
+`EnrichmentPlugin` to collect device token values with (Firebase Cloud Messaging)[]. This plugin makes it possible to collect Android's FCM and Apple's APNS device tokens.
```
oscb, line 3: Is this missing a link?

packages/plugins/plugin-device-token/README.md
```diff
@@ -0,0 +1,99 @@
...
+const segmentClient = createClient({
+  writeKey: 'SEGMENT_KEY'
+});
+
+segmentClient.add({ plugin: new DeviceTokenPlugin() });
+```
+
+### updatePermission()
+
+iOS builds require a user to give explicit permission to retrieve the APNS value. This plugin only checks to see if permission has been authorized. You will need to handle permission requests yourself. Once permission has been granted you can call the `updatePermission()` method to begin collecting the device token.
```
oscb, line 47: It is the same on Android right? We don't ask permissions so you should call the same method when you want to "refresh". Maybe we can remove the iOS part here.

packages/plugins/plugin-device-token/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "bugs": {
+    "url": "https://github.com/segmentio/analytics-react-native/issues"
+  },
+  "homepage": "https://github.com/segmentio/analytics-react-native/tree/master/packages/plugins/plugin-firebase#readme",
+  "peerDependencies": {
+    "@react-native-firebase/app": "17.x",
+    "@segment/analytics-react-native": "*"
+  },
+  "dependencies": {
+    "@react-native-firebase/messaging": "^17.3.2"
```
oscb, line 52: Can we try to make this a `peerDependency` ? What was the error you had before?

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+    this.analytics?.logger.warn(
+      'Device token only available on iOS and Android platforms'
+    );
+    return undefined;
+  }
+
+  async setDeviceToken(token: string) {
+    await this.analytics?.context.set({ device: { token: token } });
+    this.analytics?.track('Device Token Retrieved');
```
oscb, line 63: technically `track` is also `async`, maybe we should await it too.

packages/plugins/plugin-device-token/src/DeviceTokenPlugin.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  async updatePermissionStatus() {
+    const isAuthorized = await this.checkUserPermission();
+
+    if (isAuthorized) {
+      let token = await this.getDeviceToken();
+
+      if (token !== undefined) {
+        this.setDeviceToken(token);
```
oscb, line 73: This should be awaited

oscb (changes requested):

packages/plugins/plugin-device-token/README.md
alanjcharles: no we check the permission status on `configure()` regardless and if the build is android the firebase sdk automatically returns the authorized status.

oscb: yea, what I mean is this phrase:
`iOS builds require a user to give explicit permission to retrieve the APNS value.`
It is the same for Android, right? They have to give permissions and we don't request them in this library, just check. So both iOS and Android should call `updatePermission()` after they request them

packages/plugins/plugin-device-token/README.md
alanjcharles: yep lol

oscb (approved):

PR #779, opened 2023-03-10
fix: async-storage as optional peer dependency

oscb:
- Makes `@react-native-async-storage/async-storage` and `react-native-get-random-values` as peer dependencies, async-storage is now optional in both core and sovran. PR by: [@mateusz1913](https://github.com/mateusz1913)
- Remove sovran type exports. PR by: [@danychi](https://github.com/danychi)
- Fixing jest setup for tests for mocking `UUID` and `react-native-get-random-values` for all tests by default.
- Fixed the async-storage mock

Co-authored-by: Mateusz Mędrek <mateusz.medrek@netguru.com>
Co-authored-by: Daniel Alarcón Amador <danychi24@gmail.com>


alanjcharles (approved):

isaachinman:
Thanks @oscb – this is a fantastic improvement/re-organisation. Any idea when we can expect a release? Will this be a major release and/or require a minimum RN version?

oscb:
@isaachinman should be out before the end of the week! Quite possibly today or tomorrow if we can merge all fixes we want to go along with this change

Alex-McLean:
@oscb was `@segment/sovran-react-native` intentionally release as `1.0.1`, rather than `0.4.6` or even `0.5.0`?

If it is intended to be a minor or major patch, then my understanding is there is also a required change to the core library's dependencies, as it is still listed as `"@segment/sovran-react-native": "^0.4.5"` [here](https://github.com/segmentio/analytics-react-native/blob/master/packages/core/package.json).

Am currently trying to include this fix in my repo, but `analytics-react-native` will still only resolve to the old `0.4.5`.

isaachinman:
@Alex-McLean @oscb Same issue here. Setting resolutions works as a temporary fix:

```
"resolutions": {
  "@segment/sovran-react-native": ">=1.0.1"
}
```

Clearly this shows that the repo is missing any sort of functional testing in its deploy/release pipeline.

oscb:
@Alex-McLean yea, that was a mistake. It is the first release of that package from this monorepo, so `npm version` bumped the major version unexpectedly. Just released a fix in [v2.13.3](https://www.npmjs.com/package/@segment/analytics-react-native/v/2.13.3) to fix the dependency version in the core package

@isaachinman the monorepo does run functional integration tests against the example app, but it uses module resolutions to resolve and test the code to the commit being released. I'm guessing we could bump versions and deploy to a local registry, then install dependencies from there so that we could spot version bump mismatches before actually releasing to the public NPM Registry, but curious if you have any suggestions.

isaachinman:
Thanks @oscb for a quick fix!

The workflow you are describing with either a local registry, or canary versions on a remote registry, sounds like the right approach to me.

I'd highly suggest having a look at [auto](https://intuit.github.io/auto/) if you're not already aware of it.

PR #782, opened 2023-03-10
 fix: move firebase to peer dependency

alanjcharles:


packages/plugins/plugin-firebase/package.json
```diff
@@ -45,13 +45,11 @@
...
-    "@segment/analytics-react-native": "*"
-  },
-  "dependencies": {
-    "@react-native-firebase/analytics": "^15.4.0",
-    "@react-native-firebase/app": "^15.4.0"
+    "@segment/analytics-react-native": "*",
+    "@react-native-firebase/app": "^17.x"
   },
   "devDependencies": {
+    "@react-native-firebase/analytics": "^17.3.2",
```
oscb, line 52: This should be also in the `peerDependencies` right? People still need to install it. This will only install it for devs

oscb (changes requested):
single comment

packages/plugins/plugin-firebase/package.json
alanjcharles: oh, when i do that I run into the same problem I did with the messaging one (ie can't find module)

oscb: Happens if you leave it twice? As in both as a 'peerDependency' and a 'devDependency'?

PR #805, opened 2023-03-30
fix: update logPurchase conditions

alanjcharles:
- updated track events to account for Braze's new(ish) `logPurchase` behavior
-  adds ability to track purchase events when `revenue` is present
- tracks purchase events for `Order Completed` or `Completed Order` by default


oscb (changes requested):
Since this is moving the `track` code to the main plugin file I think a lot of the things i called out where/are fixed in my PR for adding more strict linting. We can merge that first or you can check my changes there and integrate into this PR.

PR #811, opened 2023-04-12
add saveDelay option for persistor

ku8ar:
We use MMKV to save the events. We don't need default 1s delay. Such config logs fewer events (when the user kills the app quickly)


oscb:
Hi @ku8ar , this is a good thing to have it exposed as an option in the client. Thanks for the suggestion.

I think it would be worth passing it on the rest of the [storages](https://github.com/segmentio/analytics-react-native/blob/master/packages/core/src/storage/sovranStorage.ts#L286), in particular the `userInfoStore` as events can have side effects on those which should also be persisted quickly in your scenario.

ku8ar:
hi @oscb , hank you for your quick response. I have added saveDelay in the rest of the places.

oscb:
@ku8ar thanks so much! lgtm

oscb:
@ku8ar this is failing to run because the types here and here are `Number` (which is a class type in TS) where it should be `number` (which is primitive).
Since this is coming from a fork I wasn't able to edit the PR directly. If you can do that I can approve this PR and merge, or I can make another PR with your changes and that fix if that's easier.

oscb (approved):
lgtm! Thanks for the contribution!

thianphat:
Hi @ku8ar, I was curious if you could provide an example of making this work with MMKV. I've tried adding the MMKV storage to storePersistor:

`storePersistor: storage`

Where `storage` is equal to `new MMKV()`

However, it's not working and of course, I am getting type errors where the methods don't align. How did you go about integrating this? Really appreciate it.

ku8ar:
@thianphat just create simple persistor and write everything as JSON:

```js
import { isNil } from 'ramda'
import { MMKV } from 'react-native-mmkv'

const storage = new MMKV({ id: 'segment' })

const storePersistor = {
  set: (key, value) => {
    storage.set(key, JSON.stringify(value))
  },
  get: (key) => {
    const mmkvValue = storage.getString(key)

    if (!isNil(mmkvValue)) {
      return JSON.parse(mmkvValue)
    }
  }
}

export default storePersistor

```

PR #820, opened 2023-04-28
fix(async-storage-persistor): Correctly import AsyncStorage 

krizzu:
We recently discovered that our Application Install events are rising exponentially. Turns out the store context was not correctly persisted, due to the wrong import on the CommonJS module of AsyncStorage.

Also, I think it'd be great to also mention in docs that users need to use store persistor so that some context events are not repeated on every App start. It's easy to miss in docs and Segment does not provide a default implementation out of the box already set up.

### DIFF
--- a/packages/sovran/src/persistor/async-storage-persistor.ts
+++ b/packages/sovran/src/persistor/async-storage-persistor.ts
@@ -7,7 +7,7 @@ let AsyncStorage: {
 
 try {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-  AsyncStorage = require('@react-native-async-storage/async-storage');
+  AsyncStorage = require('@react-native-async-storage/async-storage').default;
 } catch (error) {
   AsyncStorage = null;
 }


ltazmin:
+1, we also started seeing this problem. Nice catch @krizzu.

oscb (approved):
Thanks for catching this. 
I agree, that was a big oversight of the side effect of not having a persistor, we will add a note in there.

servercimen:
@oscb Is the sovran version 1.0.3 supposed to be available from npm? The latest version I see on the registry is 1.0.2 https://www.npmjs.com/package/@segment/sovran-react-native?activeTab=readme

efstathiosntonas:
Our events have sky rocketed:
Screenshot 2023-05-23 at 10 52 43
Screenshot 2023-05-23 at 10 52 22

niallzato:
@efstathiosntonas can you please write into friends@segment.com, so our support team can investigate the issue with your MTUs? Thank you!

efstathiosntonas:
@niallzato thanks, we’re already in an ongoing investigation with segment

PR #822, opened 2023-05-02
feat: add native anonymousId functionality

alanjcharles:
- exposes native methods to send user-generated anonymousIds across the bridge 
- adds sovran to android packages in example app


README.md
```diff
@@ -154,6 +154,44 @@ To track deep links in iOS you must add the following to your `AppDelegate.m` fi
   return YES;
 }
 ```
+### Native Anonymous ID 
+
+If you need to generate an `anonymousId` either natively or before the Analytics React Native package is initialized, you can send that value across the bridge and utilize it in the JavaScript layer. Segment does not generate this value for you, but does extend the functionality necessary to dispatch the value across the bridge. For reference, you can find a working example in the app and reference the code below:
```
oscb, line 159: Before React Native is initialized I would say, not the package. 

I would rephrase this: 
`you can send that value across the bridge and utilize it in the JavaScript layer. Segment does not generate this value for you, but does extend the functionality necessary to dispatch the value across the bridge`

to:
`you can send the anonymousId value from native code. The value has to be generated and stored by the caller. `

README.md
```diff
@@ -154,6 +154,44 @@ To track deep links in iOS you must add the following to your `AppDelegate.m` fi
...
+...
+  @Override
+  public void onCreate() {
+    super.onCreate();
+    ...
+
+  // generate your anonymousId value
+  // dispatch it across the bridge
+
+  analytics.setAnonymousId("My-New-Native-Id");
```
oscb, line 191: You need to show here that in Android you need to manually add your package so that you have the `analytics` reference.

example/android/app/src/main/java/com/example/segmentanalyticsreactnative/MainApplication.java
```diff
@@ -13,12 +13,18 @@
...
+import com.segmentanalyticsreactnative.AnalyticsReactNativeModule;
 import com.analyticsreactnativepluginadvertisingid.AnalyticsReactNativePluginAdvertisingIdPackage;
-
+import com.sovranreactnative.Sovran;
+import android.util.Log;
 
 public class MainApplication extends Application implements ReactApplication {
 
+  private AnalyticsReactNativePackage analytics = new AnalyticsReactNativePackage();
+  private Sovran sovran = new Sovran();
```
oscb, line 24: nit: Technically you don't need a reference to this as you are not using any function here directly

packages/core/android/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -1,4 +1,4 @@
-arguments=
+arguments=--init-script /var/folders/pm/jpn8gr2n7sb9v1m_mdz_bvbr0000gn/T/d146c9752a26f79b52047fb6dc6ed385d064e120494f96f08ca63a317c41f94c.gradle --init-script /var/folders/pm/jpn8gr2n7sb9v1m_mdz_bvbr0000gn/T/52cde0cfcf3e28b8b7510e992210d9614505e0911af0c190bd590d7158574963.gradle
```
oscb, line 1: We shouldn't be checking in these files (all the `android/.settings`) folder

packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativePackage.kt
```diff
@@ -10,11 +10,29 @@ import com.facebook.react.uimanager.ViewManager
 import com.facebook.react.bridge.JavaScriptModule
 
 class AnalyticsReactNativePackage : ReactPackage {
+
+    private var isInitialized = false
+    private var anonymousId = ""
```
oscb, line 15: Don't make it default to `""` as it will send it as an empty string every time this module starts.
A better idea would be to make this nullable and check if it is set to anything else than null before sending. That should save bridge calls and also prevent resetting things on the RN side (or having an explicit check for `""` which is pretty brittle)

packages/core/src/storage/sovranStorage.ts
```diff
@@ -89,6 +89,31 @@ registerBridgeStore({
...
+            ...state.userInfo,
+            anonymousId: nativeAnonymousId,
+          },
+        };
+      } else {
+        return state;
+      }
+    } else {
+      return state;
+    }
```
oscb, line 114: you can simplify with a single `return state`

oscb (changes requested):

README.md
alanjcharles: i like it

.gitignore
```diff
@@ -37,6 +37,7 @@ local.properties
 android.iml
 .settings/
 bin/
+android/.settings/
```
oscb, line 40: It is already there just below these lines, the reason why it is still adding your changes is because the files were not deleted from the repo.

You have to delete them (send a PR with the files removed) so that git doesn't track them anymore

README.md
```diff
@@ -154,6 +155,56 @@ To track deep links in iOS you must add the following to your `AppDelegate.m` fi
...
+
+...
+private AnalyticsReactNativePackage analytics = new AnalyticsReactNativePackage();
+
+...
+   @Override
+    protected List<ReactPackage> getPackages() {
+      @SuppressWarnings("UnnecessaryLocalVariable")
+      List<ReactPackage> packages = new PackageList(this).getPackages();
+      // Packages that cannot be autolinked yet can be added manually here, for
```
oscb, line 191: Remove this comment since this is the RN stuff. Add a comment:
`AnalyticsReactNative will be autolinked by default, but to send the anonymousId before RN startup you need to manually link it to store a reference to the package`

packages/core/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt
```diff
@@ -218,6 +225,24 @@ class AnalyticsReactNativeModule : ReactContextBaseJavaModule, ActivityEventList
...
   }
 
+    fun setAnonymousId(anonymousId: String) {
+     val properties = Hashtable<String, String>()
+     properties["anonymousId"] = anonymousId
+
+      if (currentActivity == null) {
+        Log.d(name, "No activity found")
+        return
+      }
```
oscb, line 235: I know you were getting an error about this before but this means that if the activity is not found it is going to drop all the calls.

Did you confirm it is actually sending the anonymousId before RN startup after this change? Or why activity was false?

oscb (changes requested):

oscb (approved):

Issue #813, opened 2023-04-15
Settings CDN proxy url


teejteej:
The analytics-js library supports setting a proxy not only for tracking, but also for the Settings CDN url.
Is there support for this in the analytics-react-native library as well?

Or is there some other way around this? For example, using an invalid writeKey, and then always supplying defaultSettings?
That wouldn't be the nicest way around it though.

Thanks!

alanjcharles:
Hi @teejteej thanks for reaching out. Can I ask what your use-case is for this? We are not against it, just trying to get more context as we haven't had requests for this in react native quite yet. Thanks!

alanjcharles:
thanks @teejteej  I'll follow up as soon as I can!

hvardhan-unth:
Addressed in PR https://github.com/segmentio/analytics-react-native/pull/834

PR #834, opened 2023-05-19
feat: add config option for cdn proxy

alanjcharles:


packages/core/src/analytics.ts
```diff
@@ -293,7 +293,17 @@ export class SegmentClient {
...
+
+    if (this.config.cdnProxy !== undefined && this.config.cdnProxy !== null) {
+      // Sets an alternative CDN host for settings retrieval. This is useful when
+      // a proxy is in use, or settings need to be queried from certain locales at
+      // all times (such as the EU). The default value is `https://cdn-settings.segment.com/v1/projects`.
+
+      settingsEndpoint = `${this.config.cdnProxy}}/${this.config.writeKey}/settings`;
+    } else {
+      settingsEndpoint = `${settingsCDN}/${this.config.writeKey}/settings`;
+    }
```
oscb, line 306: nit: you can make this a one liner
```
const prefix = this.config.cdnProxy ?? settingsCDN;
settingsEndpoint = `${prefix}/${this.config.writeKey}/settings`;
```

oscb (approved):

packages/core/src/analytics.ts
alanjcharles: #813

PR #836, opened 2023-05-24
fix: queue storage fixes

oscb:
- QueueFlushingPlugin now supports different store keys for different destination queues
- Reduce saveDelay to zero in queue to persist events immediately
- Warn if asyncStorage package is not present when using the AsyncStoragePersistor
- Added e2e tests for persistence


alanjcharles (approved):
lgtm!

PR #837, opened 2023-05-24
fix: removes Inject plugins

oscb:
Adding and storing context and userInfo is now completely handled in the client, outside of the timeline. 

This is to prevent concurrency issues in the timeline where plugins could hold off updating the userInfo, or direct operations like `reset` could cause an processing event to contain information not representative from the sequential order when it was captured.

`applyRawEventData` now is the capture moment before timeline/plugin processing starts, capturing all the data relevant to the event at a single moment.

This is the first step towards completely removing shared state.

- Removes `InjectContext` and `InjectUserInfo` plugins
- Removes `InjectTraits` example too as it is now discouraged to access shared state from any plugin. (Plugins should operate only with event data or its own state).
- Client now awaits `fetchSettings` first before triggering any lifecycle and context events. This guarantees plugins that use SegmentSettings will have them available before any events are captured. (WIP)


alanjcharles (approved):
lgtm!

PR #838, opened 2023-05-24
fix: fix flush policies reference copy, add BackgroundPolicy

oscb:
- Fixes the behaviour of `flushPolicies` to be mutually exclusive with `flushAt`/`flushInterval`. This will pevent getting the default policies when specifying only policies.
- Fixes the reference copy of the `flushPolicies` to prevent mutating the user supplied array if not destructured in the method call
- Adds tests for the `flushPolicies` from the client public methods (previously only tested in unit tests)
- Adds `BackgroundFlushPolicy` to trigger a flush when the app goes into the Background. Useful for uploading events before switching or launching a different app.


alanjcharles (approved):
lgtm!

PR #843, opened 2023-06-02
Remove isString typo

AmineKenzup:
We received millions of event errors from sentry (our quota was exceeded because of it 🙃 ), What I've found that @segment/analytics-react-native doesn't have any method called **isString**
image
image

### DIFF
--- a/packages/plugins/plugin-appsflyer/src/methods/track.ts
+++ b/packages/plugins/plugin-appsflyer/src/methods/track.ts
@@ -1,5 +1,5 @@
 import appsFlyer from 'react-native-appsflyer';
-import { isString, TrackEventType } from '@segment/analytics-react-native';
+import { TrackEventType } from '@segment/analytics-react-native';
 
 type Properties = { [key: string]: unknown };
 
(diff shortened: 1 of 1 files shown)


oscb:
There is a `isString` method in [core](https://github.com/segmentio/analytics-react-native/blob/e4b558a95e250b1b21d677e08ffeb02a4015bda6/packages/core/src/util.ts#L127)

Are you on the latest version of `analytics-react-native`?

AminexMouad:
> There is a `isString` method in [core](https://github.com/segmentio/analytics-react-native/blob/e4b558a95e250b1b21d677e08ffeb02a4015bda6/packages/core/src/util.ts#L127)
> 
> Are you on the latest version of `analytics-react-native`?
Indeed I was using an older version (2.13.4).

AmineKenzup:
> There is a `isString` method in [core](https://github.com/segmentio/analytics-react-native/blob/e4b558a95e250b1b21d677e08ffeb02a4015bda6/packages/core/src/util.ts#L127)
> 
> Are you on the latest version of `analytics-react-native`?

Indeed I was using an older version (2.13.4).

PR #844, opened 2023-06-02
Fix can not read property of undefined when identity trait value is undefined. 

shubhamdeol:
This PR fixes the issue
https://ekeekaran-ventures.sentry.io/share/issue/c97dcae32cde4f3f832e90cd363fdd23/

in older plugin we were able to send undefined values as trait but now if I will call 

```js
identify(userId, {
   enabled: undefined
})
```

error was getting thrown in this case. Fixed this passing empty string if value is undefined.


oscb (approved):

oscb (approved):

oscb:
@shubhamdeol updated your PR to fix the prettier fixes. 
All looks good to me, thanks for the contribution! I'll merge after the CI passes.

PR #847, opened 2023-06-18
feat: add property to disable segment client

AlexanderEggers:
### Summary

This PR adds a new property to disable the segment client as if the client was never set. That can be useful in the scenario when no segment related calls should be made while running in the debug mode.

This change modifies the return type of the `createClient` function which can be an undefined object now. I believe the impact should be something users can fix quickly but this feature could help people avoid sending not needed debug scenarios to segment.


jacquesdev:
Wow this will be very helpful!

PR #852, opened 2023-07-05
Add Inspector support

zikaari:
This is take 1 at adding working Inspector support. Feedback/improvements welcome

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -47,16 +47,18 @@
   },
   "homepage": "https://github.com/segmentio/analytics-react-native#readme",
   "dependencies": {
+    "@segment/inspector-core": "^2.0.2",
     "@segment/sovran-react-native": "^1",
     "deepmerge": "^4.2.2",
     "js-base64": "^3.7.2",
+    "socket.io-client": "^4.6.2",
     "uuid": "^9.0.0"
   },
   "peerDependencies": {
-    "react-native-get-random-values": "1.x",
     "@react-native-async-storage/async-storage": "1.x",
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-get-random-values": "1.x"
   },
   "peerDependenciesMeta": {
     "@react-native-async-storage/async-storage": {
@@ -66,6 +68,7 @@
   "devDependencies": {
     "@babel/core": "^7.12.9",
     "@babel/runtime": "^7.12.5",
+    "@react-native-async-storage/async-storage": "^1.0",
     "@react-native-community/eslint-config": "^2.0.0",
     "@semantic-release/changelog": "^6.0.1",
     "@semantic-release/commit-analyzer": "^9.0.2",
@@ -90,9 +93,8 @@
     "semantic-release": "^19.0.3",
     "semantic-release-monorepo": "^7.0.5",
     "ts-jest": "^27.0.7",
-    "typescript": "^4.4.4",
     "type-fest": "^3.6.1",
-    "@react-native-async-storage/async-storage": "^1.0"
+    "typescript": "^4.4.4"
   },
   "react-native-builder-bob": {
     "source": "src",
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -59,6 +59,7 @@ import {
   SegmentError,
   translateHTTPError,
 } from './errors';
+import { Inspector } from './inspector';
 
 type OnPluginAddedCallback = (plugin: Plugin) => void;
 
@@ -272,6 +273,8 @@ export class SegmentClient {
 
       // flush any stored events
       this.flushPolicyExecuter.manualFlush();
+
+      await Inspector.init();
     } catch (error) {
       this.reportInternalError(
         new SegmentError(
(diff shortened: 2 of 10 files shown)


packages/core/src/analytics.ts
```diff
@@ -272,6 +273,8 @@ export class SegmentClient {
 
       // flush any stored events
       this.flushPolicyExecuter.manualFlush();
+
+      await Inspector.init();
```
oscb, line 277: Can we make this enable only when running while on debug? You can use `this.config.debug` to check

packages/core/src/analytics.ts
```diff
@@ -415,6 +418,14 @@ export class SegmentClient {
   async process(incomingEvent: SegmentEvent) {
     const event = await this.applyRawEventData(incomingEvent);
 
+    Inspector.trace({
+      timestamp: new Date().toISOString(),
+      stage: 'triggered',
+      event: event as never,
```
oscb, line 424: `event as never` ?

packages/core/src/inspector.ts
```diff
@@ -0,0 +1,92 @@
...
+  static async init() {
+    // Inspector has already been initialized
+    if (typeof this.socket === 'object') return;
+
+    // announce the presence of this session
+    this.socket = io('http://localhost:3000', {
+      query: {
+        platform: Platform.OS,
+        sdk: 'React Native',
+        appName: await nativeModule.getAppName(),
```
oscb, line 41: It could receive this as an argument to decouple it from the NativeModules

packages/core/package.json
```diff
@@ -47,16 +47,18 @@
   },
   "homepage": "https://github.com/segmentio/analytics-react-native#readme",
   "dependencies": {
+    "@segment/inspector-core": "^2.0.2",
     "@segment/sovran-react-native": "^1",
     "deepmerge": "^4.2.2",
     "js-base64": "^3.7.2",
+    "socket.io-client": "^4.6.2",
```
oscb, line 54: My concern right now is that `socket.io` seems to be a huge library.  1MB+ in size.

Metro doesn't do Tree Shaking so it can be a lot of imports that we don't use (and possibly users won't either in prod builds)

Can we refactor this into a standalone package before publishing it?

An idea is to dispatch events at all the places where we currently do a `inspector.trace` call, the inspector can then subscribe to those events. It could retain the same APIs and `TraceData` object for simplicity.

If we refactor it into a standalone plugin lib we could even dismiss the `debug` check I mentioned in a comment below as we could refer people to only install it as `devDependency`.

packages/core/src/inspector.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    // welcome the inspector client when it joins, and get it up to speed
+    this.socket.on('inspector-joined', async (sid: string) => {
+      const send = async (method: string, args: object): Promise<boolean> => {
+        const status = (await this.socket.emitWithAck(
+          'inspector-api',
+          sid,
+          method,
+          args
+        )) as number;
```
oscb, line 53: Do we need to handle any errors in this promise?

packages/core/src/inspector.ts
```diff
@@ -0,0 +1,92 @@
...
+          if (!(await send('trace', trace))) return false;
+        }
+
+        return true;
+      };
+
+      const traces = this.queue;
+
+      const isSuccessfullBootup =
+        (await send('start', {})) && (await sendTraces(traces));
```
oscb, line 69: I'm curious as to why send the current queue here instead of as the first "prevBatch".

Is it common to experience disconnects in the socket during this step?

packages/core/src/timeline.ts
```diff
@@ -76,6 +77,15 @@ export class Timeline {
       }
     }
 
+    Inspector.trace({
+      timestamp: new Date().toISOString(),
+      stage: 'delivered',
```
oscb, line 82: Are the stage values an enum? do they show up differently in the inspector depending on the value here?

packages/core/src/timeline.ts
```diff
@@ -76,6 +77,15 @@ export class Timeline {
...
+      event: result as never,
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      id: incomingEvent.messageId!,
```
oscb, line 85: nit: I don't think this should ever ever happen, but maybe use `result` instead of `incomingEvent` as technically a plugin can change the `messageId`

packages/core/src/timeline.ts
```diff
@@ -103,6 +113,21 @@ export class Timeline {
             } else {
               await pluginResult;
             }
+
+            if ([PluginType.before, PluginType.enrichment].includes(type)) {
```
oscb, line 117: Should we also notify for `after` plugins? maybe not as an `enriched` but as a special stage?

packages/core/src/timeline.ts
```diff
@@ -103,6 +113,21 @@ export class Timeline {
...
+              Inspector.trace({
+                timestamp: new Date().toISOString(),
+                stage: 'enriched',
```
oscb, line 120: How about triggering this for all plugin types, switch the `stage` depending on the type.

As for the name of the plugin it could pull the `key` for the Destinations ([it's the only one that is enforced at the moment](https://github.com/segmentio/analytics-react-native/blob/e1dfeb8ad852350a902e4255bf7525a8af205fe6/packages/core/src/plugin.ts#L117)):

```ts
let stageMap = {
  PluginType.enrichment: "enriched",
  PluginType.before: "enriched",
  PluginType.destination: "delivered",
  PluginType.after: "after",
// ...
};
const stage = stageMap[type];

 Inspector.trace({
	timestamp: new Date().toISOString(),
	stage: stage,
    event: result as never,
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    id: event.messageId!,
    enricher: {
	 name: (type === PluginType.destination) ? plugin.key : undefined,
      type: PluginType[type],
    },
})
```

oscb (changes requested):
This is huge! 💯 

I left a few comments. The only blocker imo is about adding `socket.io` directly as a dependency to `core`. I think we should explore spinning this up as a standalone package.
Aside from that it looks great.

Great job!

PR #863, opened 2023-07-27
Add online flush policy

zikaari:


example/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "@react-native-async-storage/async-storage": "^1.15.7",
+    "@react-native-community/netinfo": "^9.4.1",
```
oscb, line 24: Since this is a native module can you add to the `readme` that they have to manually install this dependency?

In RN Native dependencies are not autolinked (only the NPM package is installed)

oscb (changes requested):
One note about the readme

Another thing to note, since the automation in this repo uses conventional commits the commit title should be: `feat: add online flush policy` so that it gets version bumped properly (important just before merging)

hvardhan-unth:
Comments resolved in new PR: #889

PR #876, opened 2023-09-08
fix: split userInfo/context stamping from raw event data

oscb:
- Splits the logic of stamping `userInfo` and `context` out from `applyRawEventData`. This is required as the raw event data should only be basic data: `timestamp` and `messageId` that can be calculated from the start, the `userInfo` and `context` should wait for the initialization to finish
  - This caused a bug where events triggered at the very start of the app, just before initialization would not contain either `userInfo` or `context`
  - Added tests

- Removed `info.ts` from the build as it is  auto generated on build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,21 +1,21 @@
 name: CI
 on:
   push:
-    branches: [ master ]
+    branches: [master]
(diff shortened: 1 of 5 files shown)


alanjcharles (approved):
lgtm!

PR #878, opened 2023-09-11
fix: switch FBSDK logTime to number

oscb:
- Changes `_logTime` in FBSDK plugin to use Unix Timestamp (in number format, non-decimal).
- Changes default to `undefined` (edge case)
- Fixes an issue in the previous implementation: `_logTime` would never be set and always relied on the default value instead of using the `event.timestamp`.
- Fixed and added tests.

### DIFF
--- a/packages/plugins/plugin-facebook-app-events/src/FacebookAppEventsPlugin.ts
+++ b/packages/plugins/plugin-facebook-app-events/src/FacebookAppEventsPlugin.ts
@@ -148,7 +148,6 @@ export class FacebookAppEventsPlugin extends DestinationPlugin {
     const safeEvent = mappedPropNames(
       event as unknown as Record<string, unknown>
     );
-    console.log(safeEvent);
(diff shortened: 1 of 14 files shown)


alanjcharles (approved):

packages/plugins/plugin-facebook-app-events/src/FacebookAppEventsPlugin.ts
```diff
@@ -147,6 +148,7 @@ export class FacebookAppEventsPlugin extends DestinationPlugin {
     const safeEvent = mappedPropNames(
       event as unknown as Record<string, unknown>
     );
+    console.log(safeEvent);
```
esdrasetrenne, line 151: log needed?

oscb: Thanks for spotting this!

PR #882, opened 2023-09-20
Add consent management and support for onetrust cmp

zikaari:
This patch implements the work laid out in [mobile consent management SDD](https://docs.google.com/document/d/19eTI6tciQWgCgIrCe6lzUZoNsdzT0HJWM3-BwpsAz3g/).

A great starting point for reviewers to check compliance with the SDD is [`OneTrust.test.ts`](https://github.com/segmentio/analytics-react-native/pull/882/files#diff-2542da281b0a1f3e8321299d3d04346fabb7bf6f14a34b814a943d3ce333e534) file.

### DIFF
--- a/packages/core/src/plugins/ConsentPlugin.ts
+++ b/packages/core/src/plugins/ConsentPlugin.ts
@@ -1,0 +1,119 @@
+import {
+  Plugin,
+  type SegmentClient,
+  type DestinationPlugin,
+  IntegrationSettings,
+  PluginType,
+  SegmentAPIIntegration,
+  SegmentEvent,
+} from '..';
+
+const SEGMENT_PREF_UPDATE_EVENT = 'Segment Consent Preference';
+
+export interface CategoryConsentStatusProvider {
+  setApplicableCategories(categories: string[]): void;
+  getConsentStatus(): Promise<Record<string, boolean>>;
+  onConsentChange(cb: (updConsent: Record<string, boolean>) => void): void;
+  shutdown?(): void;
+}
+
+/**
+ * This plugin interfaces with the consent provider and it:
+ *
+ * - stamps all events with the consent metadata.
+ * - augments all destinations with a consent filter plugin that prevents events from reaching them if
+ * they are not compliant current consent setup
+ * - listens for consent change from the provider and notifies Segment
+ */
+export class ConsentPlugin extends Plugin {
+  type = PluginType.before;
+
+  constructor(
+    private consentCategoryProvider: CategoryConsentStatusProvider,
+    private categories: string[]
+  ) {
+    super();
+  }
+
+  configure(analytics: SegmentClient): void {
+    super.configure(analytics);
+    analytics.getPlugins().forEach(this.injectConsentFilterIfApplicable);
+    analytics.onPluginLoaded(this.injectConsentFilterIfApplicable);
+    this.consentCategoryProvider.setApplicableCategories(this.categories);
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    this.consentCategoryProvider.onConsentChange(this.handleConsentChange);
+  }
+
+  async execute(event: SegmentEvent): Promise<SegmentEvent> {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if ((event as any).event === 'Segment Consent Preference') {
+      return event;
+    }
+
+    (event.context ??= {}).consent = {
+      categoryPreferences:
+        await this.consentCategoryProvider.getConsentStatus(),
+    };
+
+    return event;
+  }
+
+  shutdown(): void {
+    this.consentCategoryProvider.shutdown?.();
+  }
+
+  private injectConsentFilterIfApplicable = (plugin: Plugin) => {
+    if (this.isDestinationPlugin(plugin)) {
+      const settings = this.analytics?.settings.get()?.[plugin.key];
+
+      // FIXME: What stance should we take when this `if` block is false?
+      if (this.containsConsentSettings(settings)) {
+        plugin.add(
+          new ConsentFilterPlugin(settings.consentSettings.categories)
+        );
+      }
+    }
+  };
+
+  private handleConsentChange = async (updConsent: Record<string, boolean>) => {
+    await this.analytics?.track(SEGMENT_PREF_UPDATE_EVENT, {
+      consent: {
+        categoryPreferences: updConsent,
+      },
+    });
+  };
+
+  private isDestinationPlugin(plugin: Plugin): plugin is DestinationPlugin {
+    return plugin.type === PluginType.destination;
+  }
+
+  private containsConsentSettings = (
+    settings: IntegrationSettings | undefined
+  ): settings is Required<Pick<SegmentAPIIntegration, 'consentSettings'>> => {
+    return (
+      typeof (settings as SegmentAPIIntegration)?.consentSettings
+        ?.categories === 'object'
+    );
+  };
+}
+
+/**
+ * This plugin reads the consent metadata set on the context object and then drops the events
+ * if they are going into a destination which violates's set consent preferences
+ */
+class ConsentFilterPlugin extends Plugin {
+  type = PluginType.before;
+
+  constructor(private categories: string[]) {
+    super();
+  }
+
+  execute(event: SegmentEvent): SegmentEvent | undefined {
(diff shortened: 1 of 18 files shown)


packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -0,0 +1,119 @@
...
+    analytics.getPlugins().forEach(this.injectConsentFilterIfApplicable);
+    analytics.onPluginLoaded(this.injectConsentFilterIfApplicable);
+    this.consentCategoryProvider.setApplicableCategories(this.categories);
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    this.consentCategoryProvider.onConsentChange(this.handleConsentChange);
+  }
+
+  async execute(event: SegmentEvent): Promise<SegmentEvent> {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if ((event as any).event === 'Segment Consent Preference') {
```
zikaari, line 49: Why haven't we typed `event.event` yet?

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+  };
+
+  private handleConsentChange = async (updConsent: Record<string, boolean>) => {
+    await this.analytics?.track(SEGMENT_PREF_UPDATE_EVENT, {
+      consent: {
+        categoryPreferences: updConsent,
+      },
+    });
```
zikaari, line 83: I imagine that this event should exclusively be sent to Segment.io destination.

@silesky I checked analytics-next implementation, I saw a `track` call being made on the `analytics` instance - won't that go to all destinations?

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -0,0 +1,119 @@
...
+    super();
+  }
+
+  execute(event: SegmentEvent): SegmentEvent | undefined {
+    const preferences = event.context?.consent?.categoryPreferences;
+
+    // all categories this destination is tagged with must be present, and allowed in consent preferences
+    return this.categories.every((category) => preferences?.[category])
+      ? event
+      : undefined;
```
zikaari, line 117: @oscb I checked Timeline implementation, and the tests pass - but I want to be sure that this is the "official" way to prevent an event from advancing on.

packages/core/src/plugins/ConsentPlugin.ts
oscb: It is typed, but `SegmentEvent` is a super class, and `event` only exists in the child class: `TrackEvent`: https://github.com/segmentio/analytics-react-native/blob/527d0a0d48d7691883854c7e3583266280133dab/packages/core/src/types.ts#L41

packages/plugins/plugin-onetrust/src/OTProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+          withLatestFrom(initialCategoryStatuses$),
+          map(([accruedUpdates, initialStatuses]) => ({
+            ...initialStatuses,
+            ...accruedUpdates,
+          }))
+        );
+      }),
+      finalize(() => oneTrust.stopListeningForConsentChanges()),
+      shareReplay(1)
+    );
```
oscb, line 92: I understand OneTrust API is quite barebones and probably informed a bunch of this decision, but I find RxJS very confusing to follow.
Seems like an overkill for doing the `getConsentStatusForCategory` and `listenForConsentChanges`. It worries me a bit about maintaining it due to RxJS not being too common around this team.

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  async execute(event: SegmentEvent): Promise<SegmentEvent> {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if ((event as any).event === 'Segment Consent Preference') {
+      return event;
+    }
+
+    (event.context ??= {}).consent = {
+      categoryPreferences:
+        await this.consentCategoryProvider.getConsentStatus(),
```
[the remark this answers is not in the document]
oscb: I think in OneTrust is not too expensive though I would still suggest using a sovran store as that gives you persistence and it would let it work offline

oscb (changes requested):
I think this looks really good!
Left a few comments. Just 2 major things;
- RxJS, can we simplify? 
- Storage, let's make sure this can work offline too

Left a conversation there about the behavior that libraries should take when the consent information is not available. Might need some slight changes.

oscb (approved):

oscb:
🚢  I already approved but saw the Android Integration tests failing.
I don't see anything in your changes that could be impacting those tests, might be just a blip in Detox. The only other possibility I can think of is if Detox has an issue with NPM 18 as I see you bumped this value, although I would be surprised if it was only impacting Android.

I just put a retry run. Let's see if that clears it. Else might require some investigation.

packages/plugins/plugin-onetrust/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Integrating with Segment
+
+Interested in integrating your service with us? Check out our [Partners page](https://segment.com/partners/) for more details.
+
+## License
+
+```
+MIT License
+
+Copyright (c) 2021 Segment
```
chrisradek, line 57: Minor: Copyrights should probably be updated to 2023

packages/plugins/plugin-onetrust/package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "@segment/analytics-react-native-plugin-onetrust",
+  "version": "0.0.0",
```
chrisradek, line 3: What version will this end up getting published as?

chrisradek (commented):
Overall looks good to me! (Though I'm not very familiar with this codebase)

packages/plugins/plugin-onetrust/package.json
zikaari: Depends on what we decide while running `yarn version` command - whether we want to release it as `1.0.0` or `1.0.0-rc` or `1.0.0-alpha` etc. I don't have knowledge on what our stance is around shipping this entire feature.

packages/plugins/plugin-onetrust/README.md
zikaari: The dangers of copy and paste I see, will fix it

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -0,0 +1,134 @@
+import {
+  Plugin,
+  type SegmentClient,
```
silesky, line 3: I didn't know about this syntax!

packages/plugins/plugin-onetrust/jest.config.js
```diff
@@ -0,0 +1,16 @@
...
+  roots: ['<rootDir>'],
+  setupFiles: ['../../core/src/__tests__/__helpers__/setup.ts'],
+  testPathIgnorePatterns: ['.../../core/src/__tests__/__helpers__/'],
+  modulePathIgnorePatterns: ['/lib/'],
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest',
+  },
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  modulePaths: [compilerOptions.baseUrl],
+  moduleNameMapper: pathsToModuleNameMapper(compilerOptions.paths),
```
silesky, line 15: Didn't know about this helper!

packages/plugins/plugin-onetrust/src/OneTrust.ts
```diff
@@ -0,0 +1,9 @@
+import { ConsentPlugin } from '@segment/analytics-react-native';
+
+import { OTPublishersNativeSDK, OTCategoryConsentProvider } from './OTProvider';
+
```
silesky, line 4: Very clean!

silesky (changes requested):
LGTM 
Looks very clean 🚀

silesky (approved):

packages/plugins/plugin-onetrust/package.json
oscb: Seems like for Kotlin it is being released as 1.0.x right now
I think if we change the commit message to `feat: consent-management` ti will release it properly

PR #883, opened 2023-09-26
Update peer dependency to version 18 of Firebase

jacquesdev:
I needed this change in my project otherwise it creates a dependency issue (since I'm using rn-firebase 18.4.0). I've been using this for about two weeks now without any issues, ie all segment events are being sent as expected


mklsk (approved):

oscb (approved):
thanks for taking care of this!

jacquesdev:
Thanks @oscb! Can you confirm when a new release will be created? In this case, "@segment/analytics-react-native-plugin-firebase": "0.3.11" ?

mklsk:
@oscb would beyond appreciate an update on the upcoming release where this change will be included 🙏

PR #887, opened 2023-10-09
fix: add unknown option to current state

ashton-huxtable:
- Added a check for `currentState` to allow for the `unknown` value occasionally generated by RN per this [GH issue](https://github.com/facebook/react-native/issues/37793) in iOS. This will prevent the backgroundFlushPolicy from failing on initial launch.

### DIFF
--- a/packages/core/src/flushPolicies/background-flush-policy.ts
+++ b/packages/core/src/flushPolicies/background-flush-policy.ts
@@ -18,7 +18,7 @@ export class BackgroundFlushPolicy extends FlushPolicyBase {
       'change',
       (nextAppState) => {
         if (
-          this.appState === 'active' &&
+          (this.appState === 'active' || 'unknown') &&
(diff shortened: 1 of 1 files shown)


packages/core/src/flushPolicies/background-flush-policy.ts
```diff
@@ -18,7 +18,7 @@ export class BackgroundFlushPolicy extends FlushPolicyBase {
       'change',
       (nextAppState) => {
         if (
-          this.appState === 'active' &&
+          (this.appState === 'active' || 'unknown') &&
```
oscb, line 21: If left as right now it would use the value `unknown` and always be true. To do the intended check if would have to be as in: 

```suggestion
          (this.appState === 'active' || this.appState === 'unknown') &&
```

ashton-huxtable:
@oscb - code has been updated per your review

oscb (approved):
Thanks Ashton!

PR #894, opened 2023-11-06
Consent plugin updates and test cases

zikaari:
This work is direct derivative of @bsneed's https://github.com/segment-integrations/analytics-swift-consent/tree/main which in turn is a derivative of [consent test cases document](https://docs.google.com/document/d/1pcSWQShvyDOI-b2cx-NKc3iUqE_S40juSQK-KA6BMI8/edit).

This patch follows advisory of @bsneed and mirrors his work closely - so any amendments that follow will likely be applicable to both repositories.


oscb (approved):
really good test coverage! looks great!

PR #897, opened 2023-11-15
[bugfix] reset anonymousId properly

jboteros:
`.reset(true)` function had a bad behavior because when is resetting the anonymousId comes form `this.state.userInfo` instead of `this.userInfo`, consulting a previous version of the field.

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -562,7 +562,7 @@ export class SegmentClient {
   async alias(newUserId: string) {
     // We don't use a concurrency safe version of get here as we don't want to lock the values yet,
     // we will update the values correctly when InjectUserInfo processes the change
-    const { anonymousId, userId: previousUserId } = this.store.userInfo.get();
+    const { anonymousId, userId: previousUserId } = this.userInfo.get();
 
     const event = createAliasEvent({
       anonymousId,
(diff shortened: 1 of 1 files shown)


LucioTrucco (approved):

oscb:
@jboteros can we get more details on how to repro the issue you are bumping into here? 

`this.userInfo` just maps into `this.store.userInfo` so in theory it should be the same thing either way. I'm curious if this has fixed your issues and how to repro to understand better what's going on.

jboteros:
> @jboteros can we get more details on how to repro the issue you are bumping into here?
> 
> `this.userInfo` just maps into `this.store.userInfo` so in theory it should be the same thing either way. I'm curious if this has fixed your issues and how to repro to understand better what's going on.

Yes, that implementation fix our bug and now we have a proper reset user flow.

Basically the issue was related of the instance of the `userInfo` because you reset the `userInfo` in `this.state.userInfo` but fetch it in `this.userInfo` (at this point that instance is not update)

For now I created a patch over you package waiting for your fix and update it

Thanks!

alanjcharles:
Hi @jboteros apologies but is this still an issue? Oscar is no longer with Segment and I'm just trying to catch up on things here. Thanks!

hvardhan-unth:
Raised a PR with updated code
PR : https://github.com/segmentio/analytics-react-native/pull/950

PR #902, opened 2023-12-05
feat: RN 0.72 Upgrade

oscb:
# RN 0.72 upgrade

This is a full upgrade of the monorepo to support RN 0.72+ for testing/development.

Added another 0.73 App for testing (not working with E2E tests yet).

All devDependencies have been upgraded: Yarn, TS, Jest, ESLint, Detox, etc

Some changes:

- New `shared` private package for holding common dev utils
- E2E and Example are now split between 2 different apps
- Tests are not included in the built package anymore
- The E2E tests against release version of the app, also refactored not to use hacks for URL replacement but proxy options.

Known issues:

- Detox is not running properly in GHA. It runs locally fine.
~~- TSJest marks non-errors on individual packages (e.g. Argument of type lib.SegmentClient is not assignable to src.SegmentClient) due to configurations at individual level. Monorepo level testing works properly~~


alanjcharles (approved):
great work man!

.github/workflows/ci.yml
```diff
@@ -32,46 +40,44 @@ jobs:
...
-      - uses: actions/setup-node@v2
-        with:
-          node-version: '16'
-          cache: 'yarn'
-
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: 14.1
-
+          xcode-version: 14.3.1
```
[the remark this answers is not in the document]
oscb: I was just trying to test in the CI if the reason why the E2E couldn't run was the Xcode version.

Doesn't really change anything if we switch it up, but I will put 15 back to keep it consistent to our machines.

packages/core/src/__tests__/methods/identify.test.ts
```diff
@@ -77,7 +77,7 @@ describe('methods #identify', () => {
     });
   });
 
-  it('does not persist identity traits accross events', async () => {
+  it.only('does not persist identity traits accross events', async () => {
```
chrisradek, line 80: Looks like we left a `.only` here.

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -30,12 +30,16 @@ export interface CategoryConsentStatusProvider {
...
 
   constructor(
-    private consentCategoryProvider: CategoryConsentStatusProvider,
-    private categories: string[]
+    consentCategoryProvider: CategoryConsentStatusProvider,
+    categories: string[]
   ) {
     super();
+    this.consentCategoryProvider = consentCategoryProvider;
+    this.categories = categories;
```
chrisradek, line 42: 🤞

packages/core/src/__tests__/methods/identify.test.ts
oscb: This is exactly why I need a code review for these big changes!

chrisradek (commented):
Thanks for doing this work - had a couple questions, and the `.only` that was left in, but otherwise looks good!

packages/shared/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@segment/analytics-rn-shared",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Shared dev code between Analytics React Native projects",
+  "main": "index.js",
```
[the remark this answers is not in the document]
oscb: Right now it is just shared mocks and config files. 
Mocks get injected through the jest.config so that we don't have to repeat mocking RN and other deps everywhere.
And configs just get extended or referenced by path, hence why it isn't really a package to build right now.

chrisradek (approved):
Did upgrading detox help the e2e in ci situation?

oscb:
> Did upgrading detox help the e2e in ci situation?

@chrisradek sadly no, I'll try something else

PR #905, opened 2023-12-07
Add support for hasUnmappedDestinations

zikaari:
Implements the spec as explained [here](https://docs.google.com/document/d/13jxE3nwyn6pbUokOreJUFsfZtK7gjYq0JlPFLSUj21Y/edit)

Parity established with https://github.com/segment-integrations/analytics-swift-consent/commit/6563d95f6f1783bba4bd225d51fc849a48c7c3a5

### DIFF
--- a/packages/core/src/plugins/ConsentPlugin.ts
+++ b/packages/core/src/plugins/ConsentPlugin.ts
@@ -80,12 +80,14 @@ export class ConsentPlugin extends Plugin {
     if (this.isDestinationPlugin(plugin)) {
       plugin.add(
         new ConsentFilterPlugin((event) => {
+          const allCategories =
+            this.analytics?.consentSettings.get()?.allCategories || [];
           const settings = this.analytics?.settings.get() || {};
           const preferences = event.context?.consent?.categoryPreferences || {};
 
           if (plugin.key === SEGMENT_DESTINATION_KEY) {
-            const noneConsented = Object.values(preferences).every(
-              (consented) => !consented
+            const noneConsented = allCategories.every(
+              (category) => !preferences[category]
             );
 
             return (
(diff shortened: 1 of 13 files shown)


packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -84,14 +84,14 @@ export class ConsentPlugin extends Plugin {
...
             return (
               this.isConsentUpdateEvent(event) ||
-              !(
-                Object.values(preferences).every((consented) => !consented) &&
-                Object.entries(settings)
-                  .filter(([k]) => k !== SEGMENT_DESTINATION_KEY)
-                  .every(([_, v]) => this.containsConsentSettings(v))
-              )
+              !this.isConsentFeatureSetup() ||
+              !(noneConsented && !this.hasUnmappedDestinations())
```
zikaari, line 94: @silesky @oscb I'm tunnel visioned, can you folks help me simplify line 94? I'm sure there's a way to not have these nested negations, just can't figure how??

Note that if expression on line 91 resolves to `true`, then that means event is allowed to flow, if `false` it is blocked.

oscb: I think the `isConsentFeatureSetup` is an interesting one to have here. Cause potentially it might throw a "false" positive and let all events flow if the settings object is not loaded (f.e. you have bad connection). I guess this is more of an scenario where we should let the client control what they want to do in that scenario. Not sure really if this has been talked in Consent Management before. Anyway. We don't have to solve this yet.

I think the line `!(noneConsented && !this.hasUnmappedDestinations())` could be rewritten as `!noneConsented  || !this.hasUnmappedDestinations()`. If I'm understanding correctly the spec you want to let things flow if everything is consented or when there's at least an unmapped destination.

oscb: another idea to avoid many negations which make it harder to understand is to swap `noneConsented` to `allConsented` so that you don't have to negate.

oscb (approved):
looks good! left an idea for how to simplify the boolean logic

oscb (approved):
looks good! left an idea for how to simplify the boolean logic

PR #906, opened 2023-12-13
Relay IDFA information through consent plugin

zikaari:


packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -47,24 +47,21 @@ export class ConsentPlugin extends Plugin {
...
-        .catch((e) => {
-          throw e;
-        });
+    this.consentCategoryProvider.onConsentChange(() => {
+      this.notifyConsentChange();
+    });
+
+    let lastDeviceAttrs = analytics.context.get()?.device;
+    analytics.context.onChange((c) => {
+      const newAttrs = c?.device;
```
oscb, line 56: Is there any reason why not check the specific values inside `context.device` ? (e.g. AdTrackingEnabled, AdvertisingId)

`device` context contains a lot more values than ATT so those might change and trigger a notifyConsentChange without necessarily being an ad tracking related thing

zikaari: Actually, that was how my original implementation worked but I feared that in case more metadata becomes subject of IDFA, we might forget to also update this plugin as well.

Most fields in `device` are static and very unlikely to updated during the runtime of the app.

wdyt?

oscb: I think it's better to constraint it to the specific fields. 
It's more likely that we extend `context` with something not IDFA related and might be harder to spot that this has a side effect if it's not directly named. We also have to take into account that plugins can modify the context same as IDFA/AdvertisingID so we cannot control for what other devs are doing.

oscb (approved):

oscb (approved):
lgtm!

Issue #907, opened 2023-12-14
Braze plugin does not successfully de-duplicate Identify calls


Jack-Gill-TH:
- `analytics-react-native` version: 2.15.0
- Integrations versions (if used): @segment/analytics-react-native-plugin-braze: 0.5.4
- React Native version: 0.68.7
- iOS or Android or both? Both

**Steps to reproduce**
Make an initial call to identify with a specific set of traits
Make a second call to identify with the same set of traits

**Expected behavior**
Identify is not forwarded to braze

**Actual behavior**
Identify is forwarded to braze

---
This is because the code that checks for a duplicate payload just performs a referential equality on the two traits objects: https://github.com/segmentio/analytics-react-native/blob/0a79010b847bd36579757442aad56242f816bdae/packages/plugins/plugin-braze/src/BrazePlugin.tsx#L92

I think it's fairly unlikely that this code will be called each time with the same object reference, so this should be updated to perform some kind of deep equality.

oscb:
Closing. Fix is released in [v0.6.1](https://www.npmjs.com/package/@segment/analytics-react-native-plugin-braze/v/0.6.1)

PR #912, opened 2024-01-15
Niall/braze debounce update

niallzato:
Our current logic for comparing the last seen traits and the current traits does not actually compare values. Our current logic to compare the traits is just checking the referential equality:

`      
this.lastSeenTraits?.traits === event.traits
`

Referential equality can be determined with equality operators such as strict equality (===) or coercive equality (==) and also by using [Object.is](http://object.is/)() functions, but determining deep equality is tricky as the objects can be nested.

Customer issue: https://github.com/segmentio/analytics-react-native/issues/907

### DIFF
--- a/packages/plugins/plugin-braze/src/BrazePlugin.tsx
+++ b/packages/plugins/plugin-braze/src/BrazePlugin.tsx
@@ -83,56 +83,36 @@ export class BrazePlugin extends DestinationPlugin {
     return undefined;
   };
 
-  private isObject(object: any):boolean {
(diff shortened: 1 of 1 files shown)


alanjcharles (commented):
lgtm! Can you write a test for it in [identify.test.ts](https://github.com/segmentio/analytics-react-native/blob/master/packages/plugins/plugin-braze/src/methods/__tests__/identify.test.ts) ?

oscb (approved):

oscb (approved):

PR #916, opened 2024-02-07
fix: preserve pending events when initialization doesn't complete (offline)

oscb:
- Preserve pending events triggered before client can initialize in case f.e. the App is starting on a slow network/offline and user closes the app
- Change Sovran to support Arrays not wrapped into an object
- Tests
- Remove unused e2e/jest stuff from example app


alanjcharles (approved):
lgtm!

PR #917, opened 2024-02-09
fix: update gradle files for 0.73 support

### LINKED ISSUE #915: Please add support for React Native 0.73 and Expo SDK 50
Please add support for React Native 0.73 and Expo SDK 50 which now defaults to Java 17 for the build environment.

- `analytics-react-native` version: 2.15.0
- Integrations versions (if used): 
- React Native version: 0.73
- iOS or Android or both? both

Currently builds are failing due to Java version being pinned at 11:
image

oscb:
Current lib should just work with 0.73 but taking preliminary steps to keep up to date based on the work specified [here](https://github.com/react-native-community/discussions-and-proposals/issues/671)
And in particular on the revised outputs of [create-react-native-library CLI](https://github.com/callstack/react-native-builder-bob/blob/fc8744092501ba00fd11bdb69e1ecb46fbb63c0f/packages/create-react-native-library/templates/native-common/android/build.gradle)

- Updated build.gradle and gradle.properties to match new CLI outputs
- Added new AndroidManifest.xml without namespace
- Added namespace in build.gradle only if Gradle 7.3+

Fixes #915


packages/core/android/build.gradle
```diff
@@ -1,40 +1,63 @@
 buildscript {
   // Buildscript is evaluated before everything else so we can't use getExtOrDefault
-  def kotlin_version = rootProject.ext.has('kotlinVersion') ? rootProject.ext.get('kotlinVersion') : project.properties['AnalyticsReactNative_kotlinVersion']
+  def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["AnalyticsReactNative_kotlinVersion"]
 
   repositories {
     google()
-    jcenter()
+    mavenCentral()
```
chrisradek, line 7: Was curious about this - found https://developer.android.com/build/jcenter-migration which explains why you needed to do this :smile:
(And I guess it also just matches the new template)

chrisradek (approved):
Looks good to me - did you already verify this doesn't break things for react-native < 0.73?

oscb:
> Looks good to me - did you already verify this doesn't break things for react-native < 0.73?

Yes, the E2E actually tests against 0.72
I cannot build personally anything less than 0.72 due to Xcode MacOS incompatibility. Apple is also removing the capability to submit apps not built with Xcode 15 in a few months so people will be upgrading to 0.72 at least.

Also of note that we're not bumping any major versions or using any particular features of Java/gradle so people should just be able to override with lower numbers if needed just as before

packages/core/android/build.gradle
oscb: I was actually surprised that we hadn't done this before! 
I remember it used to be very noisy in analytics-android/kotlin

PR #929, opened 2024-03-18
fix(plugin-facebook): Fix Dedup issues in Facebook Plugin

oscb:
- Adds `event_id` from `messageId` to the FB events to support deduplication
- Adds a `_logTime` unit test to verify the E notation error (no-repro)
- Bumps `FBSDK-Next` to latest


alanjcharles (approved):
lgtm!

PR #942, opened 2024-04-12
fix: remove idfv fallback for idfa plugin

ashton-huxtable:
- removes fallback to `idfv` value for `context.device.advertisingId` when a user asks app not to track

### DIFF
--- a/packages/plugins/plugin-idfa/ios/AnalyticsReactNativePluginIdfa.swift
+++ b/packages/plugins/plugin-idfa/ios/AnalyticsReactNativePluginIdfa.swift
@@ -16,11 +16,11 @@ class AnalyticsReactNativePluginIdfa: NSObject {
     ) -> Void {
         if #available(iOS 14, *) {
             ATTrackingManager.requestTrackingAuthorization { status in
-                let idfa = status == .authorized ? ASIdentifierManager.shared().advertisingIdentifier.uuidString : nil
-                if let authorizedIDFA = idfa {
+            if status == .authorized {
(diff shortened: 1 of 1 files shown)


packages/plugins/plugin-idfa/ios/AnalyticsReactNativePluginIdfa.swift
```diff
@@ -16,21 +16,28 @@ class AnalyticsReactNativePluginIdfa: NSObject {
...
         if #available(iOS 14, *) {
             ATTrackingManager.requestTrackingAuthorization { status in
-                let idfa = status == .authorized ? ASIdentifierManager.shared().advertisingIdentifier.uuidString : self.fallbackValue
-                resolve([
-                    "adTrackingEnabled": status == .authorized,
-                    "advertisingId": idfa!,
-                    "trackingStatus": self.statusToString(status)
-                ])
+                let idfa = status == .authorized ? ASIdentifierManager.shared().advertisingIdentifier.uuidString : nil
+                if let authorizedIDFA = idfa {
```
oscb, line 20: nit: You could refactor this to check only for status in the if then assign the value to avoid the ternary:

```swift
if status == .authorized {
  let idfa = ASIdentifierManager.shared().advertisingIdentifier.uuidString
  //...
}
//...
```

ashton-huxtable: refactored per this suggestion

alanjcharles (approved):

PR #943, opened 2024-04-16
fix: startup flush policy should send restored events

oscb:
Problem reported that StartupFlushPolicy wouldn't send the pending upload events from the previous app launch. 

Caused by concurrency: FlushPolicies get initialized before and without awaiting for storage restoration. Futhermore the Queue for upload is managed by the SegmentDestination plugin so there's no explicit way to await for that queue to be restored from the base client itself.

Solution:
- Move the flush policy initialization later in the client: after plugins are initialized but before replaying buffered events
- Refactor StartupFlushPolicy to mark `shouldFlush` outside of `start` (this is to be able to handle it with the `manualFlush`)
- Refactor `QueueFlushingPlugin` and `SegmentDestination` to await for queue restoration and settings load when a flush is triggered. This guarantees that if the flush is triggered before the plugins are fully initialized (previous session events loaded) it will await for those operations to complete before uploading any events.

### DIFF
--- a/packages/core/src/plugins/QueueFlushingPlugin.ts
+++ b/packages/core/src/plugins/QueueFlushingPlugin.ts
@@ -3,6 +3,8 @@ import type { SegmentClient } from '../analytics';
 import { defaultConfig } from '../constants';
 import { UtilityPlugin } from '../plugin';
 import { PluginType, SegmentEvent } from '../types';
+import { createPromise } from '../util';
+import { ErrorType, SegmentError } from '../errors';
 
 /**
  * This plugin manages a queue where all events get added to after timeline processing.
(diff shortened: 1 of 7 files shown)


alanjcharles (approved):

packages/core/src/plugins/QueueFlushingPlugin.ts
```diff
@@ -60,6 +73,19 @@ export class QueueFlushingPlugin extends UtilityPlugin {
    * Calls the onFlush callback with the events in the queue
    */
   async flush() {
+    // Wait for the queue to be restored
+    try {
+      await this.isRestored;
+    } catch (e) {
+      // If the queue is not restored before the timeout, we will notify but not block flushing events
```
didiergarcia, line 80: Any chance we get another type of error here that is not related to timeouts? Should those be handled any differently?

packages/core/src/plugins/SegmentDestination.ts
```diff
@@ -89,6 +96,12 @@ export class SegmentDestination extends DestinationPlugin {
   configure(analytics: SegmentClient): void {
     super.configure(analytics);
 
+    // If the client has a proxy we don't need to await for settings apiHost, we can send events directly
+    // Important! If new settings are required in the future you probably want to change this!
```
didiergarcia, line 100: Should this maybe be a config option is we might need to change it in the future?

packages/core/src/util.ts
```diff
@@ -232,3 +232,23 @@ export function deepCompare<T>(a: T, b: T): boolean {
 
   return true;
 }
+
+export const createPromise = <T>(
```
didiergarcia, line 236: Maybe a more descriptive name like `createTimerPromise` or `createTimeoutPromise` ?

didiergarcia (approved):

packages/core/src/plugins/SegmentDestination.ts
oscb: A config in the client? I would prefer us to control it. If we end up changing it is because some code was added in the library itself to use those new settings that require awaiting. For people on old versions of the lib where this behavior is still around it shouldn't be a problem.

packages/core/src/util.ts
oscb: it used to be `createTimeoutPromise` but then I realized I made the timeout optional.

packages/core/src/plugins/QueueFlushingPlugin.ts
oscb: Not really, any other issue during restoration will be handled elsewhere. JS is notoriously limited in error handling.

PR #944, opened 2024-04-17
Include updated Expo instructions for deep link tracking

leopic:
Adds missing instructions of how to setup Automating Deep Link tracking for iOS, under Expo.

The current set of instructions assume Objective-C and a non-volatile AppDelegate.


oscb:
This is great! Thanks for adding this

PR #946, opened 2024-05-13
Upgrade braze react native sdk to 10.x

hvardhan-unth:
Updated the braze react native sdk from 5.x to 10.x (latest version).

The latest version of braze react native sdk supports react native version 0.73 which is not supported in barze sdk 5.x.

Tested from local system using emulator.

### DIFF
--- a/packages/plugins/plugin-braze/package.json
+++ b/packages/plugins/plugin-braze/package.json
@@ -46,11 +46,11 @@
   },
   "homepage": "https://github.com/segmentio/analytics-react-native/tree/master/packages/plugins/plugin-braze#readme",
   "peerDependencies": {
-    "@braze/react-native-sdk": "^8.x",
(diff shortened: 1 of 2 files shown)


alanjcharles:
@hvardhan-unth I think these tests are failing bc `yarn.lock` has been modified and this is a `public` pull request. Can you fix that? If not, it shouldn't block merging but it's better to make sure all of the CI tests are passing if we can help it. Thanks!

hvardhan-unth:
@alanjcharles Please check the PR again as I have removed the unnecessary changes from the yarn.lock file.
The changes that are present in yarn.lock file are necessary as we are upgrading the version of braze-react-native sdk.

alanjcharles (approved):

alanjcharles:
thanks!

PR #959, opened 2024-06-13
fix: refactor ConsentPlugin and OneTrust Plugin

alanjcharles:
- removes `OneTrustPlugin()` 
- adds update method to `ConsentPlugin` to get consent settings 
- updates documentation to provide implementation clarity


didiergarcia (approved):

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -65,16 +70,24 @@ export class ConsentPlugin extends Plugin {
...
+            await this.consentCategoryProvider.getConsentStatus(),
+        },
+      };
+      return event;
+    }
 
-    return event;
+    if (this.consentStarted === false) {
+      this.queuedEvents.push(event);
+      console.log('Event queued', event);
```
silesky, line 87: Console?

alanjcharles: erg. i swear i always miss one

silesky (approved):

packages/core/src/plugins/ConsentPlugin.ts
wenxi-zeng: may want to remove this debug log?

packages/core/src/plugins/ConsentPlugin.ts
```diff
@@ -151,6 +163,19 @@ export class ConsentPlugin extends Plugin {
       throw e;
     });
   }
+
+  public start() {
```
wenxi-zeng, line 166: when is this function getting called? didn't see it anyway except in unit test. maybe need to be invoked via sovran state changes?

wenxi-zeng (dismissed):

packages/core/src/plugins/ConsentPlugin.ts
alanjcharles: yep done thanks!

packages/core/src/plugins/ConsentPlugin.ts
alanjcharles: it's called by the user when they want to start processing queued events (could depend on when their CMP is initialized) and it's how swift/kotlin work. This way, we queue events until they want to be processed. Otherwise, events like Application Lifecycle events don't get stamped with consent preferences.

alanjcharles: described in the `readme` if you're curious!

wenxi-zeng: ah. I see. should've read the file. in that case, we may want to bound the queue to avoid OOM

silesky (approved):

wenxi-zeng (approved):

PR #969, opened 2024-07-01
Fix/libweb 1755

hvardhan-unth:
Issue : 
@segment/analytics-react-native-plugin-braze - Invariant Violation: new NativeEventEmitter() requires a non-null argument

Fix : 
Added the required dependencies inside sovran package.json file.


alanjcharles (approved):

alanjcharles:
@hvardhan-unth approved the code changes but the CI is failing 3 different suites. We need to fix that before we can merge

Issue #961, opened 2024-06-17
Limit batch size


matthieuMay:
When batching too much events, we often hit the 500kb limit of batch API.

Inspecting the events I realized that a lot of data sent is redundant (integrating context, _metadata, integrations)....

Reading the [documentation of batch API](https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/#batch) it seems that integrations and context should be extracted which drastically reduce the payload size.

patch-package to realize this is : 

```diff
diff --git a/node_modules/@segment/analytics-react-native/src/api.ts b/node_modules/@segment/analytics-react-native/src/api.ts
index a3048da..41e6099 100644
--- a/node_modules/@segment/analytics-react-native/src/api.ts
+++ b/node_modules/@segment/analytics-react-native/src/api.ts
@@ -9,12 +9,17 @@ export const uploadEvents = async ({
   url: string;
   events: SegmentEvent[];
 }) => {
+  const context = events.find((event) => !!event.context)?.context;
+  const integrations = events.find((event) => !!event.integrations)?.integrations;
+  let sentEvents = events.map(({context,integrations, ...event}) => (event));
   return await fetch(url, {
     method: 'POST',
     body: JSON.stringify({
-      batch: events,
+      batch: sentEvents,
       sentAt: new Date().toISOString(),
       writeKey: writeKey,
+      context: context,
+      integrations: integrations,
     }),
     headers: {
       'Content-Type': 'application/json; charset=utf-8',
```

hvardhan-unth:
Hey @matthieuMay 
Thanks for the good suggestion. 
Please raise a PR for the implementation you have mentioned here so that our team can review and consider it further.

PR #971, opened 2024-07-03
feat: Refactor API uploadEvents function

matthieuMay:
as suggested here: https://github.com/segmentio/analytics-react-native/issues/961#event-13380371229

This greatly limits the size of the payloads sent 

We have been using it in production for 3 weeks without any problem

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -9,12 +9,17 @@ export const uploadEvents = async ({
   url: string;
   events: SegmentEvent[];
 }) => {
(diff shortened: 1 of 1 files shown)


sunitaprajapati89:
@matthieuMay : Thanks for your suggestion. But the changes you proposed makes the code out of sync from other platforms which is not a good idea. Also, regarding [documentation of Batch API](https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/#batch), I would look into this again. 

Feel free to raise PR if you have other enhancements in mind.

PR #972, opened 2024-07-08
Fix/libweb 1767

hvardhan-unth:
Issue : 
Mixpanel - Missing increment properties method

Fix : 
Update the mixpanel key and added the constructor for using increment property, also added first_name & last_name to the userTraits.


hvardhan-unth:
Not required.

Issue #973, opened 2024-07-10
Queue timeout before completed restoration


zgordon02:
- `analytics-react-native` version:
^2.19.2
- Integrations versions (if used):
N/A
- React Native version:
0.72.1 AND 0.74.3
- iOS or Android or both?
BOTH

![Screenshot 2024-07-10 at 1 04 04 PM](https://github.com/segmentio/analytics-react-native/assets/90853783/071b8be9-6834-4af7-aa8d-026512fe0252)

**Steps to reproduce**
Send various screen and track events

Here is an example screen event I am sending:
`SCREEN event saved {"name": "TEST_SCREEN", "properties": {"_environment": "LOCAL", "_otaVersion": "N/A"}, "type": "screen"}`

**Expected behavior**
No warning
**Actual behavior**
This warning appears multiple times per session. However, the events are shown properly in the segment console

zgordon02:
Important to note that I have left almost all options to their default:

```
    segmentClient.current = createClient({
      writeKey: environment?.segmentKey,
      trackAppLifecycleEvents: true,
    });
```

mayfairr:
++ Also getting this.

umairmaqbool3:
I am also getting same warnings... 
![checkerror](https://github.com/user-attachments/assets/a3f50e5d-16fd-49d4-b8f7-e482cf855a53)

mcoeur:
Same here with version 2.19.2.
Adding a comment to get notifications on this thread 👌

seanadkinson:
Also seeing these warnings spewed in my console.

MarkKravchuk:
Same here, any ideas how to suppress or silence these warnings?

zgordon02:
> Same here, any ideas how to suppress or silence these warnings?

I have not tried this, but the code should be something like this:

```
import { LogBox } from 'react-native';
// ...
LogBox.ignoreLogs(["An internal error occurred: Error: Queue timeout before completed restoration"]);

```

leethree:
I think this is caused by this recent change https://github.com/segmentio/analytics-react-native/pull/943

The QueueFlushingPlugin has a default timeout of 500ms and it will fail if it's not intialized within that time. But in our testing it sometimes takes 2 seconds for the entire thing to be ready. So it will block every events from being sent.

zgordon02:
> So it will block every events from being sent.

FWIW I have not noticed events dropped while seeing this warning. It appears to be functioning normally

[...]

PR #984, opened 2024-08-06
fix: increase default timeout and change warning to info

alanjcharles:
- increases the `restoreTimeout` to `1000` 
- changes warning to an info message to be less intrusive in console


alanjcharles:
resolves #973

wenxi-zeng (approved):

PR #992, opened 2024-08-23
Extend peer dependency support to v19 and v20 of react native Firebase

adkenyon:
We're getting several peer dependency issues when using the firebase plugin with rn-firebase v19. I've updated the peer dependency declarations to work for v19 (what we're running in prod despite peer dependency warnings) and for v20 (typescript compiler seems to accept and inspection of firebase API seems to show no breaking changes with analytics)

This is effectively a rehash of #883


adkenyon:
@alanjcharles

alanjcharles:
addressed in #996

Issue #982, opened 2024-08-02
Error: Promise timed out


lucasoliveiraw00:
I'm facing a "Promise timed out" error in my application. Is anyone else having this issue?
  
![Captura de Tela 2024-08-01 às 21 22 52](https://github.com/user-attachments/assets/e0ac0e88-e4f8-45f0-8f13-9cc5acd2a32e)
![Captura de Tela 2024-08-01 às 21 29 34](https://github.com/user-attachments/assets/782b95f0-df1c-40de-b1b6-b0767128db82)

- `analytics-react-native` version: `2.19.2`
- Integrations versions (if used):
- React Native version: `0.73.8`
- iOS or Android or both? Android

kip-west:
> I'm facing a "Promise timed out" error in my application. Is anyone else having this issue?
> 
[...]

Issue #954, opened 2024-06-07
[2.19.2] Possible Unhandled Promise Rejection (id: 0): Error: Promise timed out


efstathiosntonas:
I get this on app start: 

```
Possible Unhandled Promise Rejection (id: 0):
Error: Promise timed out
Error: Promise timed out
```

and later on: 

```
Promise Rejection Handled (id: 0):
 WARN    This means you can ignore any previous messages of the form "Possible Unhandled Promise Rejection" with id 0.
```

* https://github.com/segmentio/analytics-react-native/blob/e079c71958d8f57cdc7bf4f1131bd2ce4d483284/packages/core/src/util.ts#L236

- `analytics-react-native` version: 2.19.2
- Integrations versions (if used):
- React Native version: 0.74.2
- iOS or Android or both? Both

**Steps to reproduce**

**Expected behavior**
don't know
**Actual behavior**
throws promise timeout

efstathiosntonas:
It appears that this is happening only on 2.19.2, downgrading to 2.19.1 does not show the rejection any more.

gitwilliam:
It may be unrelated, but I was getting the same error mention in the OP and also: `Queue timeout before completed restoration`. Downgrading to 2.19.1 resolved both of them.

alanjcharles:
hi folks thanks for the report- will prioritize this and follow up asap.

alanjcharles:
@efstathiosntonas @gitwilliam At present, the only PR I can find related to this issue is #943. From what I can tell in my investigation so far, these errors on app launch are expected and described in detail in that PR mentioned above. The change was introduced to prevent flushing before plugins (including SegmentDestination) are initialized. I can reproduce the warnings you are seeing but I am not experiencing any event loss. 

Please let me know if you are experiencing event loss or any other unintended/regressive behavior and we can go from there.

jvgeee:
@alanjcharles No event loss, but I'm getting these errors every few seconds when running my app (not just at app launch). Please have them as opt-in through a config flag or something, it's clogging up my debug terminal!

<img width="620" alt="image" src="https://github.com/segmentio/analytics-react-native/assets/2121331/d2052258-7fd6-48e6-ad85-b1f3e1a2fac4">

efstathiosntonas:
@alanjcharles same as @jvgeee, It’s clogging the terminal and it’s really annoying since our minds automatically tries to find out what is going on when we see a promise rejection happening 😂

[...]

PR #999, opened 2024-09-12
Fix: add conditional to queue restoration error reporting

alanjcharles:
- adds `timeoutWarned` boolean to prevent multiple warnings for the same error
- reports internal error to `analytics` for customizable action(s)
- adds conditional and message for successful flush if initial flush failed due to queue restoration timeout


packages/core/src/plugins/QueueFlushingPlugin.ts
```diff
@@ -75,12 +76,30 @@ export class QueueFlushingPlugin extends UtilityPlugin {
     // Wait for the queue to be restored
     try {
       await this.isRestored;
+
+      if (this.timeoutWarned === true) {
```
wenxi-zeng, line 80: I guess this statement is trying to prevent the timeout message from keep popping up? two questions (forgive me if they are naive):
1. not sure how `this.isRestored` works, but looks the catch block is intended for`this.isRestored`. so if we have this if statement after the await, will it ever be triggered? because if there is an exception, it goes straight to the catch block. 
2. should we have `timeoutWarned` be reset to false somewhere? maybe when the flush actually succeed? otherwise the value may not be true for some cases.

wenxi-zeng (approved):

packages/core/src/plugins/QueueFlushingPlugin.ts
alanjcharles: not at all! 
1. yes it will be triggered on a successfully resolved promise of `this.isRestored`
2. good call- added it

MichaelGHSeg (approved):

alanjcharles:
#954 #982

meabed:
I get this error now:
image

```
 WARN  An internal error occurred:  [Error: Queue restoration timeout]
```

PR #1015, opened 2024-10-23
fix: advertising id being sent for Android application installed events using analytics-react-native-plugin-advertsing-id

thesunita:
Fix for - Advertising Id not being sent for Android Application Installed events using analytics-react-native-plugin-advertising-id #995

### DIFF
--- a/packages/plugins/plugin-advertising-id/src/AdvertisingIdPlugin.ts
+++ b/packages/plugins/plugin-advertising-id/src/AdvertisingIdPlugin.ts
@@ -5,6 +5,7 @@ import {
   getNativeModule,
   ErrorType,
   SegmentError,
+  SegmentEvent
 } from '@segment/analytics-react-native';
 
 import { Platform, NativeModule } from 'react-native';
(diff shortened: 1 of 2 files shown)


packages/plugins/plugin-advertising-id/src/AdvertisingIdPlugin.ts
```diff
@@ -48,6 +52,16 @@ export class AdvertisingIdPlugin extends Plugin {
...
   }
 
+  execute(event: SegmentEvent){
+
+    if (this.advertisingId === undefined) {
+      this.queuedEvents.push(event);
+    }else{
+      return event;
+    }
+    return;
```
cgadam, line 62: What does it means in this context to return `undefined`?  What if there's other plugin that is meant to be executed after this plugin? Does this means that the other plugin won't get anything to process? Will the other plugin indeed eventually get something once we execute `process` for this event?

cgadam, line 62: A more clear why for this `if` could also be:

```
if (this.advertisingId === undefined) {
  this.queuedEvents.push(event);
  return;
}
return event;
```

PR #1017, opened 2024-10-24
fix: Prevent potential crash in Android SDK 35 by handling nullable versionName

dmitry-blackwave:
# Overview

image

This PR fixes a potential crash in Android SDK 35 by ensuring that versionName, which is now nullable, is safely converted to a string. This prevents crashes when versionName is null during app initialization.


alanjcharles (approved):

alanjcharles:
thank you @dmitry-blackwave

PR #1024, opened 2024-11-11
fix: events not being sent when adTrackingEnabled is false

thesunita:
fix for events not being sent when adTrackingEnabled is false


alanjcharles (approved):

ignaciomendeznole:
Thanks for working on this 💟

PR #1022, opened 2024-10-31
Fix: Type mismatch: inferred type is String? but String was expected

edapess:
e: file:///..../node_modules/@segment/analytics-react-native/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt:128:30 Type mismatch: inferred type is String? but String was expected

PR #1027, opened 2024-11-13
Fix for error: Type mismatch: inferred type is String? but String was expected

wamry:
`node_modules/@segment/analytics-react-native/android/src/main/java/com/segmentanalyticsreactnative/AnalyticsReactNativeModule.kt:128:30 Type mismatch: inferred type is String? but String was expected`


wamry:
duplicate of another https://github.com/segmentio/analytics-react-native/pull/1022

PR #1043, opened 2025-01-31
fix: for Unable to configure Custom Domain for React-Native

sunitaprajapati89:
fix: for Unable to configure Custom Domain for React-Native

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,24 +118,26 @@ You must pass at least the `writeKey`. Additional configuration options are list
   
 ### Client Options
 
-| Name                       | Default   | Description                                                                                                                                    |
-| -------------------------- | --------- | -----------------------------------------------------------------------------------------------------------------------------------------------|
-| `writeKey` **(REQUIRED)**  | ''        | Your Segment API key.                                                                                                                          |
-| `collectDeviceId`          | false    | Set to true to automatically collect the device Id.from the DRM API on Android devices.                                           |
-| `debug`                    | true\*    | When set to false, it will not generate any logs.                                                                                              |
-| `logger`                   | undefined | Custom logger instance to expose internal Segment client logging.                                                                            |
-| `flushAt`                  | 20        | How many events to accumulate before sending events to the backend.                                                                            |
-| `flushInterval`            | 30        | In seconds, how often to send events to the backend.                                                                                           |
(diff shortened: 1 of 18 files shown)


wenxi-zeng:
## unblock e2e tests
the following should unblock the e2e the tests:
1. update the url in [this block](https://github.com/segmentio/analytics-react-native/blob/6153e20eb75d120193d1161fdee5ec85a3c74341/examples/E2E/App.tsx#L49-L56) to match the changes you made (i.e. `/events` to `v1/b`, `/settings` to `/projects/${writekey}/settings`).
2. do the same for [this block](https://github.com/segmentio/analytics-react-native/blob/6153e20eb75d120193d1161fdee5ec85a3c74341/examples/E2E/e2e/mockServer.js#L19-L35)

**NOTE that you need provide the write key to the endpoints**

## best practice
however, your changes hardcoded the path/endpoints to every domain, which disables the user to fully customize the url to their own endpoints and this is a breaking change. the suggest fix would be:
1. add a new flag in config, something like `useSegmentEndpoints` and default it to false. (so we don't break existing customer who use their own path/endpoints`
2. update the your code to append Segment's path/endpoints if and only if `useSegmentEndpoints` is set to true. otherwise, use the one user provided as is
4. now you can revert the changes you made on the mock server (mentioned in [unblock e2e tests](##unblock-e2e-tests) section)
5. add unit tests to test against `useSegmentEndpoints`

wenxi-zeng (commented):
LGTM. just address the only comment and update README and it's ready to go

efstathiosntonas:
UNBELIEVABLE!!!

we spent 5 hours to track this down. 

The generated endpoint is `https://api.segment.io/v1/b/b` which returns 404. 

We have missed 5 days of events and thousands of dollars loss.

efstathiosntonas:
i just can't fucking believe this. How this could have slipped? How such a serious bug went through the tests?

wenxi-zeng:
@efstathiosntonas really sorry for all the inconvenience it has caused.  @sunitaprajapati89 is working on the fix. in the meanwhile, please pass `https://api.segment.io/v1/b` as the proxy in the config or rollback to the previous version. the unit test only checks if the url ends with the correct endpoints, which makes this issue slipped away. we will update it to catch issues like this.

wenxi-zeng:
hi @efstathiosntonas the issue should be fixed in `2.20.5`. please have a try and let us know if you still experience issues.

PR #1052, opened 2025-03-11
fix: remove remaining void promises from appsflyer plugin

lennartschoch:
This PR is following up on #1025 - in that PR, I removed some usages of `void` Promises in the AppsFlyer plugin that weren't properly triggered in my project. I missed some of them in the same project, so following up here.

Let me know if you have any questions about this!


alanjcharles (approved):
thanks!

PR #1053, opened 2025-03-11
feat: add callbacks to appsflyer plugin

lennartschoch:
This PR passes the three callbacks used in the appsflyer plugin out of the plugin as props.
[Appsflyer docs](https://dev.appsflyer.com/hc/docs/rn_unifieddeeplink) mention that the `onDeepLink` callback must be configured before initializing the SDK and as far as my testing goes, the callback cannot be defined more than once per SDK instance.

By adding props for the used callbacks, it's possible to use the callback logic used by the plugin and still specify custom deep link handlers.

Let me know if I'm missing something or if there is a better or easier way to solve this!


alanjcharles (approved):
thanks again!

PR #1059, opened 2025-03-25
fix: example syntax is invalid

Atomic71:
In js/ts, if you want an arrow function to return an object in an inline manner, it needs to be surrounded by parenthesis.


alanjcharles (approved):
thanks!

PR #1067, opened 2025-04-25
feat: Ability to programmatically disable tracking

sunitaprajapati89:
feat: Ability to programmatically disable tracking

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -486,15 +486,14 @@ export class SegmentClient {
   async process(incomingEvent: SegmentEvent, enrichment?: EnrichmentClosure) {
     const event = this.applyRawEventData(incomingEvent);
     event.enrichment = enrichment;
-    if (this.enabled.get() === true) {
-      if (this.isReady.value) {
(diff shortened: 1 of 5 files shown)


packages/core/src/analytics.ts
```diff
@@ -476,12 +486,15 @@ export class SegmentClient {
   async process(incomingEvent: SegmentEvent, enrichment?: EnrichmentClosure) {
     const event = this.applyRawEventData(incomingEvent);
     event.enrichment = enrichment;
-
-    if (this.isReady.value) {
-      return this.startTimelineProcessing(event);
+    if (this.enabled.get() === true) {
```
wenxi-zeng, line 489: it's cleaner to do the following to avoid deep nested branches
```
if (this.enabled.get() === false) {
    // return ...
}
if (this.isReady.value) {
...
```

wenxi-zeng (approved):

packages/core/src/analytics.ts
sunitaprajapati89: Done

PR #1090, opened 2025-08-14
Fix race condition causing events to be lost during initialization

bytesoverflow:
This PR fixes a critical race condition that causes events to be lost when they are tracked while the Segment client is processing pending events during initialization.

## Problem
Events tracked during the app startup phase were being lost. When the Segment client initializes, it processes any events that were tracked before it was ready (pending events). 
However, due to a race condition, new events arriving during this processing period were incorrectly saved as "pending" and never sent.

### Reproduction Steps
1. Track events immediately when a React Native app starts
2. Track additional events within the first 2-4 seconds
3. Observe that only a subset of events are sent to Segment

## Root Cause
The `isReady` flag was being set AFTER the `onReady()` function completed processing pending events. This created a window where:
1. Pending events are being processed
2. New events arrive and check `isReady.value` 
3. Since `isReady.value` is still `false`, they get saved as pending
4. But the pending processor has already run, so these events are never sent

## Solution
Set the `isReady` flag to `true` BEFORE processing pending events, not after. This ensures any events arriving during initialization go directly to the queue.


abueide:
fixed in [v2.21.4](https://github.com/segmentio/analytics-react-native/releases/tag/%40segment%2Fanalytics-react-native-v2.21.4)

PR #1110, opened 2025-12-24
feat: Waiting plugin addition

sunitaprajapati89:
add waiting plugin interface.

- any plugin implement this interface will pause event processing when it's added to analytics
- any plugin implement this interface can pause and resume event processing as needed

### DIFF
--- a/packages/core/src/plugins/__tests__/Waiting.test.ts
+++ b/packages/core/src/plugins/__tests__/Waiting.test.ts
@@ -456,10 +456,7 @@ describe('WaitingPlugin', () => {
     await client.running.set(true);
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const processPendingSpy = jest.spyOn(
-      client as any,
-      'processPendingEvents'
-    );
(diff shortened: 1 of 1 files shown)


packages/core/src/analytics.ts
```diff
@@ -465,17 +476,24 @@ export class SegmentClient {
         settings
       );
     }
-
-    if (!this.isReady.value) {
+    console.log('!this.isReady.value', !this.isReady.value);
```
abueide: we should clean up & remove debug console.logs

packages/core/src/plugins/__tests__/Waiting.test.ts
```diff
@@ -0,0 +1,338 @@
...
+  });
+
+  test('test resume after timeout', async () => {
+    const client = new SegmentClient({
+      config: baseConfig,
+      logger: getMockLogger(),
+      store,
+    });
+
+    client.isReady.value = true;
```
abueide: client.isReady shouldn't manually be set, it should be set automatically after analytics.init is called on the client, or does the testing harness skip the client initialization?

abueide (changes requested):

PR #1125, opened 2026-02-18
refactor(devbox): use hosted devbox plugin from segment-integrations

abueide:
Refactor to use the react native devbox plugin hosted at https://github.com/segment-integrations/devbox-plugins


abueide:
wait till devbox plugins repo is released

PR #1297, opened 2026-06-27
fix(idfa): return promise from getTrackingAuthorizationStatus

### LINKED ISSUE #1070: IDFA plugin to expose getTrackingAuthorizationStatus response
## Summary
Can the `getTrackingAuthorizationStatus` function return the `IdfaData` so that we can decide if we can show the permission modal by calling `requestTrackingPermission`?

## Proposed Solution
It looks like the IdfaData is returned from with in the `.then`, but the whole promise is not returned from the function. Can the return be added?

https://github.com/segmentio/analytics-react-native/blob/master/packages/plugins/plugin-idfa/src/IdfaPlugin.tsx#L63

SYoder1:
The `getTrackingStatus` function in the IDFA plugin calls `getTrackingAuthorizationStatus` which returns a promise of the current tracking state, but getTrackingStatus never returned it.

Closes https://github.com/segmentio/analytics-react-native/issues/1070

### DIFF
--- a/packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
+++ b/packages/plugins/plugin-idfa/src/IdfaPlugin.tsx
@@ -60,7 +60,7 @@ make additional tracking decisions based on the user response
   }
 
   getTrackingStatus() {
-    getTrackingAuthorizationStatus()
+    return getTrackingAuthorizationStatus()
       .then((idfa: IdfaData) => {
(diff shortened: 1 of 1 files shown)


SYoder1:
@abueide if you have a free second I would love your feedback on this

abueide:
@SYoder1 looks good to me, thanks!

PR #1298, opened 2026-06-27
fix(deps): switch react-native-get-random-values to optional

SYoder1:
`react-native-get-random-values` is required as a peer dependency in both the core and sovran packages. This is used for UUID v4 generation. 

I am using `react-native-quick-crypto` in my app and already polyfill the `crypto.getRandomValues` function globally. The peer dependency forces me to install `react-native-get-random-values` even though it no-op on setup.

**On the same topic:**
I have found in my testing that the UUID generation done on the native side is about 40% faster then using the JS `uuid` package. Looking back a past PRs and issues, it seems that it has bounced around a few times from being a pure native function to then using `uuid` and a polyfill because issues while debugging.

I wonder if there is a middle ground where it checks for the best solution at run time, without needing to use a slower solution all the time.

Example: `crypto.randomUUID()` -> if native module, generate uuid  -> use polyfill and `uuid` package.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -68,6 +68,9 @@
   "peerDependenciesMeta": {
     "@react-native-async-storage/async-storage": {
       "optional": true
+    },
+    "react-native-get-random-values": {
+      "optional": true
     }
   },
   "engines": {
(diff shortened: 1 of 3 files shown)


SYoder1:
@abueide I know you just update the peer version of `react-native-get-random-values` but I would Love your thoughts on making it optional

abueide:
added with a small change in #1310

SYoder1:
Thank you!!

PR #1299, opened 2026-07-02
fix(sovran): remove unused shell-quote runtime dep

sunitaprajapati89:


samjoffe:
Note that the version of shell-quote we carry has a CVE.

PR #1306, opened 2026-07-24
ci: dependency curation fixes + devbox/nix CI toolchain (Artifactory-only)

abueide:
## Summary
- Real curation-blocked package fixes from the new audit tooling in #1305: eliminated npm@10.9.8, sigstore@3.1.0, and both stale pacote versions by overriding the old @semantic-release/npm@^11.0.0 chain that pulled them in; pinned tar/baseline-browser-mapping/electron-to-chromium and ws to exact, already-aged versions instead of floating caret overrides that kept re-resolving to whatever's newest (and thus perpetually <48h old).
- browserslist remains genuinely blocked - confirmed Xray evaluates its own declared transitive ranges against the live registry, not our lockfile, so no version pin fixes it; it's also foundational (Babel, react-native-builder-bob) so it can't be removed. Needs a curation policy exception from whoever owns twilio.jfrog.io.
- New: routes CI's toolchain through devbox/nix, resolving everything through the SDK build lane's Artifactory instance only - mirrors the pattern proven in segmentio/ajs-renderer's nix-based Buildkite migration (patched devbox honoring `DEVBOX_NIX_BINARY_CACHE` instead of the hardcoded cache.nixos.org), now pointed at the brand-new `virtual-nix-thirdparty`/`virtual-debian-thirdparty` repos (twilio-internal/artifactory-cloud-twilio-config#11, merged yesterday). Replaces `actions/setup-node` + `corepack` with `devbox run --pure` against devbox.json's existing `ci:install`/`lint`/`build`/`typecheck`/`test`/`ci:commitlint` scripts.

## Test plan
- [ ] First real integration of virtual-nix-thirdparty/virtual-debian-thirdparty by any consumer - expect iteration on repo URL shape, auth format, and Ubuntu codename matching
- [ ] `devbox version` succeeds in the bootstrap step
- [ ] `devbox run --pure ci:install` resolves node/yarn and installs dependencies entirely from Artifactory
- [ ] lint/build/typecheck/test pass through devbox
- [ ] release.yml intentionally not touched yet - proving this out on ci.yml first


abueide:
Deferring the devbox/nix CI migration - the Bearer-only token + broken virtual-debian-thirdparty repo need a server-side fix first. Scoping back to #1305, which already has just the curation-audit tooling + dependency fixes, since the priority right now is getting a release out.

PR #1326, opened 2026-09-08
fix(core): don't log full event payloads or deep-link tokens at info level

sunitaprajapati89:
Track/screen/identify/group/alias/deep-link event logging previously
passed the entire SegmentEvent (userId, traits, properties, and the raw
deep-link URL) to logger.info. The default Logger only disables itself
when NODE_ENV === 'production', which many RN release bundles don't set
for JS, and any custom logger or logger.enable() call re-exposed this
data to logcat/device logs and log-capture SDKs.

- Log only non-sensitive metadata (type, messageId, name) at info level
  by default.
- Add an explicit `debugPayloads` config flag to opt into full-payload
  logging, gated behind a startup warning about the PII/token exposure.
- Redact the deep-link URL's query string before it is ever logged,
  even when debugPayloads is enabled.
- Log the processed event (with its real messageId) instead of the
  pre-processed one, so the metadata log is actually useful for
  correlation.

### DIFF
--- a/packages/core/src/analytics.ts
+++ b/packages/core/src/analytics.ts
@@ -80,6 +80,23 @@ import { WaitingPlugin } from './plugin';
 
 type OnPluginAddedCallback = (plugin: Plugin) => void;
 
+// Deep link URLs can carry secrets/tokens in their query string, so strip it before
+// the full event is logged (only reachable when `debugPayloads` is enabled).
+const redactDeepLinkUrl = (event: SegmentEvent): SegmentEvent => {
+  if (
+    event.type === EventType.TrackEvent &&
+    event.event === 'Deep Link Opened' &&
+    typeof event.properties?.url === 'string'
+  ) {
+    const [urlWithoutQuery] = event.properties.url.split('?');
+    return {
+      ...event,
+      properties: { ...event.properties, url: urlWithoutQuery },
+    };
+  }
+  return event;
+};
+
 export class SegmentClient {
   // the config parameters for the client - a merge of user provided and default options
   private config: Config;
@@ -296,6 +313,12 @@ export class SegmentClient {
 
     // set up tracking for lifecycle events
     this.setupLifecycleEvents();
+
+    if (this.config.debugPayloads === true) {
+      this.logger.warn(
+        'debugPayloads is enabled: full event payloads (which may include PII and deep-link tokens) will be logged at info level. Do not enable this in production.'
+      );
+    }
   }
 
   // Watch for isReady so that we can handle any pending events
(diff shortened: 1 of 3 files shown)


packages/core/src/analytics.ts
```diff
@@ -80,6 +80,23 @@ import { WaitingPlugin } from './plugin';
...
 
+// Deep link URLs can carry secrets/tokens in their query string, so strip it before
+// the full event is logged (only reachable when `debugPayloads` is enabled).
+const redactDeepLinkUrl = (event: SegmentEvent): SegmentEvent => {
+  if (
+    event.type === EventType.TrackEvent &&
+    event.event === 'Deep Link Opened' &&
+    typeof event.properties?.url === 'string'
+  ) {
+    const [urlWithoutQuery] = event.properties.url.split('?');
```
didiergarcia: `split('?')` only removes the query string, but mobile OAuth callbacks put the token in the **fragment**:

```
myapp://callback#access_token=eyJhbGci...
```

That survives redaction and is logged verbatim under `debugPayloads` — the leak this PR is meant to close. Token-in-path (`myapp://reset/AbC123`) leaks too.

```suggestion
    const [urlWithoutQuery] = event.properties.url.split(/[?#]/);
```

Also worth considering an allowlist of deep-link properties instead of redacting `url` alone: the other keys spread in from `deepLinkData` (`referring_application`, plus anything the host app adds) are logged untouched.

packages/core/src/analytics.ts
```diff
@@ -600,7 +623,38 @@ export class SegmentClient {
       });
 
       void this.process(event);
-      this.logger.info('TRACK (Deep Link Opened) event saved', event);
+      this.logEventSaved('TRACK (Deep Link Opened)', event);
```
didiergarcia: This logs the pre-process event while `process()`'s result is discarded on the line above (`void this.process(event)`). `applyRawEventData` returns a new object rather than mutating its input, so `messageId` here is always `undefined` — the line reads `{ type: 'track', name: 'Deep Link Opened', messageId: undefined }`.

Every other call site was converted to log the processed event; this one wasn't, and it's the path where correlating an event to a delivered payload matters most.

```suggestion
      const processedEvent = await this.process(event);
      this.logEventSaved('TRACK (Deep Link Opened)', processedEvent ?? event);
```

That needs the enclosing method to await — if the fire-and-forget `void` was deliberate here, capturing the promise result instead would work.

packages/core/src/analytics.ts
```diff
@@ -677,8 +731,8 @@ export class SegmentClient {
       properties: options,
     });
 
-    await this.process(event, enrichment);
-    this.logger.info('SCREEN event saved', event);
+    const processedEvent = await this.process(event, enrichment);
+    this.logEventSaved('SCREEN', processedEvent ?? event);
```
didiergarcia: `processedEvent ?? event` collapses two different outcomes into "event saved". `process()` returns `undefined` when analytics is disabled (`analytics.ts:565`) and when a before/enrichment plugin drops the event (`Timeline.process`, `timeline.ts:64` — e.g. consent gating). In both cases we still log `SCREEN event saved`, and since the fallback is the un-processed event the metadata degrades to `{ type, messageId: undefined }` — it goes empty exactly when someone is debugging why their event vanished.

Prefer distinguishing them:

```ts
if (processedEvent === undefined) {
  this.logger.info('SCREEN event dropped', this.getEventLogMetadata(event));
  return;
}
this.logEventSaved('SCREEN', processedEvent);
```

Same shape at 749, 763, 777 and 792 — probably worth folding the `undefined` handling into `logEventSaved` itself.

packages/core/src/types.ts
```diff
@@ -137,6 +137,13 @@ export type Config = {
   writeKey: string;
   debug?: boolean;
   logger?: DeactivableLoggerType;
+  /**
+   * By default, the SDK only logs non-sensitive event metadata (type, messageId, event/name)
+   * at info level. Set this to log the full event payload instead, which can include PII
```
didiergarcia: "log the full event payload **instead**" — but `logEventSaved` emits both lines: metadata unconditionally, then the payload. The new test asserts exactly that (`expect(logger.info).toHaveBeenCalledTimes(2)`).

```suggestion
   * at info level. Set this to also log the full event payload, which can include PII
```

didiergarcia (changes requested):

didiergarcia (changes requested):
A few changes to capture edge cases

MichaelGHSeg (approved):

packages/core/src/analytics.ts
sunitaprajapati89: done

packages/core/src/analytics.ts
sunitaprajapati89: Done

packages/core/src/analytics.ts
sunitaprajapati89: done

packages/core/src/types.ts
sunitaprajapati89: Done

